PR #5, opened 2022-10-07
Clean up repository & fix workflows

### LINKED ISSUE #1: Clean up newly initialized Python API

### LINKED ISSUE #2: Fix workflows so they succeed

### LINKED ISSUE #8: Enable proper release workflow
The publish/release workflow will currently not work, as the `tripper` name is not available on PyPI.
An issue has been opened (pypa/pypi-support#2286) to request a transfer of ownership for this name.

CasperWA:
Fixes #1 
Fixes #2 

This PR is an initial "clean up" of the files under `tripper` as well as a code base update according to the pre-commit hooks.
Furthermore, this ensures the CI workflows pass.

---

Fixes #8 

The latest commits now enables releasing the package to PyPI under the name `tripperpy`, which should (hopefully) be updated to `tripper` once the issue pypa/pypi-support#2286 has been resolved.


CasperWA:
Issues #4, #6, and #7 have been opened to follow up on changes introduced by this PR and general information found in the code base.

docs/overview.md
```diff
@@ -145,4 +136,5 @@ Further development
     - Apache Jena Fuseki
     - Allegrograph
     - Wikidata
-* Add ontological validation of physical dimension to Triplestore.mapsTo().
+
+- Add ontological validation of physical dimension to Triplestore.mapsTo().
```
jesper-friis, line 140: Already done

jesper-friis, line 140: Already supported
```suggestion
```

tests/test_literals.py
```diff
@@ -1,34 +1,56 @@
...
-assert l4.datatype == XSD.integer
-assert l4.value == 42
-assert l4.n3() == f'"42"^^{XSD.integer}'
-
-l5 = Literal(42, datatype=float)
-assert l5.lang is None
-assert l5.datatype == XSD.double
-assert l5.value == 42.0
-assert l5.n3() == f'"42"^^{XSD.double}'
+
```
jesper-friis, line 2: I have difficulties to see the benefits of putting all tests into functions.

The old script is easier to read and so much easier to play around with during development or debugging in an good interactive interpreter like ipython.

tripper/backends/ontopy.py
```diff
@@ -51,73 +51,86 @@ def __init__(
     def triples(self, triple: "Triple") -> "Generator":
         """Returns a generator over matching triples."""
 
-        def to_literal(o, d):
-            """Returns a literal from (o, d)."""
-            if isinstance(d, str) and d.startswith("@"):
-                lang, datatype = d[1:], None
+        def to_literal(object_, datum):
```
jesper-friis, line 54: s, p, o are standard variable names when you are dealing with triples. Writing them out just makes the code lines longer and less readable. 

The `d` stands for "datatype" not "datum". The notation for language strings are e.g. "@en" for an English language string.

tripper/backends/ontopy.py
```diff
@@ -51,73 +51,86 @@ def __init__(
...
         abb = (
-            None if s is None else self.onto._abbreviate(s),
-            None if p is None else self.onto._abbreviate(p),
-            None if o is None else self.onto._abbreviate(o),
+            None if (subject) is None else self.onto._abbreviate(subject),
+            None if (predicate) is None else self.onto._abbreviate(predicate),
+            None if (object_) is None else self.onto._abbreviate(object_),
         )
-        for s, p, o in self.onto._get_obj_triples_spo_spo(*abb):
+        for subject, predicate, object_ in self.onto._get_obj_triples_spo_spo(*abb):
```
jesper-friis, line 68: Good practice is to keep the line length less than 79 characters, since longer lines are difficult to read on a standard 80 character wide terminal.

tripper/backends/ontopy.py
```diff
@@ -135,29 +148,32 @@ def parse(self, source=None, location=None, data=None, format=None,
...
                 if filename:
                     os.remove(filename)
 
         else:
-            raise ValueError(
-                "either `source`, `location` or `data` must be given"
-            )
-
-    def serialize(self, destination=None, format='turtle', **kwargs):
+            raise ValueError("either `source`, `location` or `data` must be given")
```
jesper-friis, line 169: Please keep line length below 79 characters

tripper/triplestore.py
```diff
@@ -686,14 +730,13 @@ def _add_function_fno(self, func, expects, returns, base_iri):
         self.add((func_iri, RDF.type, FNO.Function))
         self.add((func_iri, FNO.expects, parlist))
         self.add((func_iri, FNO.returns, outlist))
-        if doc:
-            self.add((func_iri, DCTERMS.description, en(doc)))
+        if inspect.getdoc(func):
```
jesper-friis, line 733: This is slightly inefficient, since it calls `inspect.getdoc(func)` twice...

tests/test_literals.py
CasperWA: It's to achieve proper test isolation to not have unexpected setups occurring, resulting in unintended and opaque environments when testing certain aspects, unknowingly performing faulty tests, since you're not specifically testing a functionality in a _fully known_ environment. Any failing tests will be difficult to debug and understand, more so the later in the script they appear.

The tests should be minimalistic enough for you to _not_ have to create breakpoints, but rather understand what part of the code is failing straight away. This is the point of unit testing as I understand it.
However, a functionality in the code base might depend on another part of the code base (most likely), in which case a proper debugging environment is necessary and should be something we should strive towards and force ourselves to implement.
Writing code without a proper debugging setup is like driving a car without a steering wheel.

It is true the pytest framework has a specific way of working and running. And I'm happy to consider another modern framework if another is suggested. However, I think it may be faster for you to learn this system than it is for everyone to learn a new one? But it might indeed be time well invested and spent.

If you crave the usage of script-based testing locally, I suggest to create _proper_ example scripts that can be used in the documentation simultaneously as for your local testing.
These can for example be placed under `/docs/examples/`.

tripper/backends/ontopy.py
CasperWA: The fact the I didn't know what `d` stands for speaks to the difficulty of using such variable names.
I will not be the only one not knowing this, nor knowing what `s`, `p`, and `o` stands for.

However, what we can do (since I also don't like having to add an underscore to `object`) is to write the naming convention into the module doc-string and include it in each of the function/method doc-strings as well (just copy-pasting). And also add `s`, `p`, and `o` as accepted variable names to the pylint configuration.

I would still prefer to _not_ use `d`, but instead use `datatype`.

tripper/backends/ontopy.py
CasperWA: "Standard" is _very_ subjective in this context.
Most modern coding environments and terminals support larger/more variable terminal windows.
Furthermore, this project is not implementing a terminal-specific API (like a command-line interface), which is less of a reason to specifically support terminals.

I instead fully support the more modern take on line lengths for more easy readability in modern code editors. Something that has been investigated and determined to be a line-length of 88 characters, which is a compromise line-length.

tripper/backends/ontopy.py
CasperWA: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

tripper/triplestore.py
CasperWA: I'm betting on clever Python caching here. But you are correct.
My goal here was to minimize the number of local variables.

tests/test_literals.py
daniel-sintef: I strongly agree with Casper. Pytest is a pretty standard framework these days. (And it is not that fundamentally different from the standard library unittest https://docs.python.org/3/library/unittest.html).  I have used breakpoints all the time and it is not *that* difficult. Agree as well that the purpose of tests is to test the code, it is not a set of user-friendly tutorials, which should be made in a separate example session.

tripper/backends/ontopy.py
daniel-sintef: To quote "Clean Code" regarding single-variable names:

> Single-letter names and numeric constants have a particular problem in that they are not
> easy to locate across a body of text.
> 
> One might easily grep for `MAX_CLASSES_PER_STUDENT`, but the number 7 could be more
> troublesome. Searches may turn up the digit as part of file names, other constant definitions,
> and in various expressions where the value is used with different intent. It is even
> worse when a constant is a long number and someone might have transposed digits,
> thereby creating a bug while simultaneously evading the programmer’s search.
> 
> Likewise, the name e is a poor choice for any variable for which a programmer might
> need to search. It is the most common letter in the English language and likely to show up
> in every passage of text in every program. In this regard, longer names trump shorter
> names, and any searchable name trumps a constant in code.
> 
> My personal preference is that single-letter names can ONLY be used as local variables
> inside short methods. **The length of a name should correspond to the size of its scope**
> 
> If a variable or constant might be seen or used in multiple places in a body of code,
> it is imperative to give it a search-friendly name. Once again compare:
> 
> ```
> for (int j=0; j<34; j++) {
> s += (t[j]*4)/5;
> }
> ```
> 
> to 
> 
> ```
> int realDaysPerIdealDay = 4;
> const int WORK_DAYS_PER_WEEK = 5;
> int sum = 0;
> for (int j=0; j < NUMBER_OF_TASKS; j++) {
> int realTaskDays = taskEstimate[j] * realDaysPerIdealDay;
> int realTaskWeeks = (realdays / WORK_DAYS_PER_WEEK);
> sum += realTaskWeeks;
> }
> ```
> 
> Note that sum, above, is not a particularly useful name but at least is searchable. The
> intentionally named code makes for a longer function, but consider how much easier it
> will be to find `WORK_DAYS_PER_WEEK `than to find all the places where 5 was used and filter
> the list down to just the instances with the intended meaning.

daniel-sintef: "Code Complete" cites interesting emperical studies: 

> Gorla, Benander, and Benander found that the effort required to debug a program was minimized when variables had names that averaged 10 to 16 characters (1990). Programs with names averaging 8 to 20 characters were almost as easy to debug. **The guideline doesn’t mean that you should try to make all of your variable names 9 to 15 or 10 to 16 characters long. It does mean that if you look over your code and see many names that are shorter, you should check to be sure that the names are as clear as they need to be.**

> Are short variable names always bad? No, not always. When you give a variable a short
> name like i, the length itself says something about the variable—namely, that the variable is a scratch value with a limited scope of operation.
> 
> A programmer reading such a variable should be able to assume that its value isn’t
> used outside a few lines of code. When you name a variable i, you’re saying, “This variable is a run-of-the-mill loop counter or array index and doesn’t have any significance
> outside these few lines of code.”
> 
> A study by W. J. Hansen found that longer names are better for rarely used variables or
> global variables and shorter names are better for local variables or loop variables

CasperWA: Indeed - but in certain domains I also agree with @jesper-friis that using "standard" conventions makes things easier. In the ontology/triple domain, `s`, `p`, and `o` are frequently used as short-names for subject, predicate, and object (the three parts of a triple).
Hence, when they come together, it should ("should" being the key word here) be clear what is meant for other coders within the same context domain. For those not in the same context domain or just learning, we can support them by writing this convention in doc-strings (for the module, a dedicated documentation page or in function/method/class doc-strings).

However, this domain convention arguments falls apart for the `d` parameter in my view, hence my comment above.

Coming back to using `s`, `p`, and `o` "when they come together", raises the issue of: How "degraded" does this togetherness have to be before we start using their full names?
I.e., for the methods that take a subject and returns the predicates and objects for said subject - is this togetherness broken? What about the methods that take two out of three and provides the last breaking the togetherness?

daniel-sintef: This is a good/interesting point to think about. What is the **scope** of these s, p, o variables? Do they frequently show up in small functions, limited loops, short methods, etc.? If so, and especially if used consistently, and in line with 'the standards' of the field, it can make sense to give them shorter names. Ad-hock labeling data/datatype/datum as d, probably not

tests/test_literals.py
jesper-friis: I would normally argue to use a standard framework. It is not that I don't understand pytest or are able to use it (like VS Code), but I have difficulties to see the real value it provides - except for test isolation, that is actually a good argument. There are also some powerful test fixtures, which might be nice when you are writing the tests, but insane if you try to understand a test someone else has written. Documentation of which conftest.py file or builtin pytest module that implements the fixture may mitigate this. But so far I haven't seen any such documentation.

On the other hand, I see when I am co-developing with others, how powerful it is for quick checks during development, to write your test script in parallel with the code and run it in a good interactive interpreter in a terminal next to your text editor. You can't do that with pytest test functions. 

The suggestion of Casper with an example script in docs/examples/ may be a compromise. That may even be a good way to force you to think about and good use case and user friendly interface for the library while writing the code.

tripper/backends/ontopy.py
jesper-friis: > I would still prefer to _not_ use `d`, but instead use `datatype`.

I bye that one, I has nothing against replacing `d` with `datatype` as long as we keep s, p, o.

jesper-friis: > This is a good/interesting point to think about. What is the **scope** of these s, p, o variables? Do they frequently show up in small functions, limited loops, short methods, etc.? If so, and especially if used consistently, and in line with 'the standards' of the field, it can make sense to give them shorter names. Ad-hock labeling data/datatype/datum as d, probably not

s, p, o are very frequently used when you are dealing with RDF triples. Writing d for datatype is an abbreviation (heavily used in owlready2, though).

tripper/backends/ontopy.py
jesper-friis: Hehe, reference against reference. For me short lines counts for readability (especially in an 80 character terminal).

jesper-friis (commented):
I think the style discussion starts to converge... the best would be if it is possible to find time for quickly agree on the best practices we want to follow.

tripper/backends/ontopy.py
CasperWA: > (...)
> Short line length and use of symbols counts for readability, at least for people with a small tendency of dyslexia. What do you think is most readable
> 
> ```
> E = m * c**2
> ```
> 
> or
> 
> ```
> Energy equals mass times speed of light in vacuum squared
> ```

In this particular example I would prefer the latter, simply due to the inherent semantics provided. That's what's mainly missing when shortening the variables. But a middle-ground would be most preferred for sure.

But your point is taken. Still, care should be taken to write as readable code as possible, by as many developers as possible, within the small sub-set of the overlapping domain knowledge that exists.

CasperWA: > > This is a good/interesting point to think about. What is the **scope** of these s, p, o variables? Do they frequently show up in small functions, limited loops, short methods, etc.? If so, and especially if used consistently, and in line with 'the standards' of the field, it can make sense to give them shorter names. Ad-hock labeling data/datatype/datum as d, probably not
> 
> s, p, o are very frequently used when you are dealing with RDF triples. Writing d for datatype is an abbreviation (heavily used in owlready2, though).

Owlready2's Python syntactics should not be repeated outside Owlready2. It is _not_ a model one should try to achieve or immitate.

CasperWA:
> I think the style discussion starts to converge... the best would be if it is possible to find time for quickly agree on the best practices we want to follow.

This might not be "quickly" done, but we have indeed started a wiki document internally to outline these best practices. Please see our internal chat thread.

PR #9, opened 2022-10-09
Documented return value of the Triplestore.query() and added a test for it

jesper-friis:
This ended up including:
- documented return value of Triplestore.query() method
- improved documentation of arguments of query() and update() methods
- corrected return value of the query() method (for rdflib backend) and added tests
- ensure that the tests runs again after the package was renamed


tripper/interface.py
```diff
@@ -56,10 +56,27 @@ def serialize(self, destination=None, format='xml', **kwargs)
         """
 
     def query(self, query_object, **kwargs)
```
CasperWA, line 58:
```suggestion
    def query(self, query_object, **kwargs):
```

tripper/interface.py
```diff
@@ -56,10 +56,27 @@ def serialize(self, destination=None, format='xml', **kwargs)
...
+
+        Parameters:
+            query_object: String with the SPARQL query.
+            kwargs: Additional backend-specific keyword arguments.
+
+        Returns:
+            List of tuples of IRIs for each matching row.
+        """
 
     def update(self, update_object, **kwargs)
```
CasperWA, line 69:
```suggestion
    def update(self, update_object, **kwargs):
```

tripper/interface.py
```diff
@@ -56,10 +56,27 @@ def serialize(self, destination=None, format='xml', **kwargs)
...
+        Parameters:
+            query_object: String with the SPARQL query.
+            kwargs: Additional backend-specific keyword arguments.
+
+        Note:
+            This method is intended for INSERT and DELETE queries.  Use
+            the query() method for SELECT queries.
+        """
 
     def bind(self, prefix: str, namespace: str)
```
CasperWA, line 81:
```suggestion
    def bind(self, prefix: str, namespace: str):
```

tripper/backends/rdflib.py
```diff
@@ -115,13 +115,32 @@ def serialize(
...
+        Parameters:
+            query_object: String with the SPARQL query.
+            kwargs: Keyword arguments passed to rdflib.Graph.query().
+
+        Returns:
+            List of tuples of IRIs for each matching row.
+
+        """
+        rows = self.graph.query(query_object=query_object, **kwargs)
+        return [tuple(str(v) for v in row) for row in rows]
```
CasperWA, line 130:
```suggestion
        return [
            tuple(str(v) for v in row)
            for row in self.graph.query(query_object=query_object, **kwargs)
        ]
```

CasperWA (changes requested):

tripper/backends/rdflib.py
jesper-friis: What is the benefit of this change? I feel that the original code is slightly easier to read and two lines shorter.

CasperWA: I found this more readable, hence the suggested change. I have no issues if you prefer the other way - I just prefer not making variables unnecessarily just to make variables, and reading directly where each row is taken from seems more readable to me. But again, it's just a suggested change, not a necessary one.

CasperWA (changes requested):
There are still some EOL, function-defining colons missing in the `interface.py` file.
The one I can see is for the `namespaces` method.

CasperWA (approved):
👍 Squash+merge at will

PR #11, opened 2022-10-09
sparqlwrapper backend

jesper-friis:
Added initial drat implementation of a SPARQLWrapper backend.

In order to test the backend, This PR builds on top on PR #9.


CasperWA (approved):

tripper/triplestore.py
```diff
@@ -168,7 +168,7 @@ def serialize(
         self._check_method("serialize")
         return self.backend.serialize(destination=destination, format=format, **kwargs)
 
-    def query(self, query_object, **kwargs) -> "List[Tuple[str, ...]]":
+    def query(self, query_object, **kwargs) -> "List[Triple]":
```
jesper-friis, line 171: This is actually not correct. The return value of `query()` depends on the sparql query. It just returns a list of tuples for each match. The tuple is not a triple but contains as many fields as the query asks for.

CasperWA: so the original return value is correct?

PR #25, opened 2022-10-23
Refactored triplestore module

### LINKED ISSUE #3: Refactor triplestore.py (triplestore module API)
The `triplestore/triplestore.py` file should be refactored and an updated Python API.

jesper-friis:
Closes #3 

Also fixed some bugs and added the functions parse_literal() and split_iri()

### DIFF
--- a/tests/test_literals.py
+++ b/tests/test_literals.py
@@ -3,7 +3,7 @@
 
 def test_string() -> None:
     """Test creating a string literal."""
-    from tripper.triplestore import Literal
+    from tripper import Literal
 
     literal = Literal("Hello world!")
     assert literal == "Hello world!"
@@ -17,7 +17,7 @@ def test_string() -> None:
 
 def test_string_lang() -> None:
     """Test creating a string literal with a set language."""
-    from tripper.triplestore import Literal
+    from tripper import Literal
 
     literal = Literal("Hello world!", lang="en")
     assert literal.lang == "en"
@@ -28,7 +28,7 @@ def test_string_lang() -> None:
 
 def test_en() -> None:
     """Test creating a string literal through `en()`."""
-    from tripper.triplestore import en
+    from tripper import en
 
     literal = en("Hello world!")
     assert literal.n3() == '"Hello world!"@en'
@@ -36,7 +36,7 @@ def test_en() -> None:
 
 def test_integer() -> None:
     """Test creating an integer literal."""
-    from tripper.triplestore import XSD, Literal
+    from tripper import XSD, Literal
 
     literal = Literal(42)
     assert literal.lang is None
@@ -47,10 +47,63 @@ def test_integer() -> None:
 
 def test_float_through_datatype() -> None:
     """Test creating a float literal from an int through datatype."""
-    from tripper.triplestore import XSD, Literal
+    from tripper import XSD, Literal
 
     literal = Literal(42, datatype=float)
     assert literal.lang is None
     assert literal.datatype == XSD.double
     assert literal.value == 42.0
     assert literal.n3() == f'"42"^^{XSD.double}'
+
+
+def test_split_iri() -> None:
+    """Test parse n3-encoded literal value."""
+    from tripper import DCTERMS, RDFS, split_iri
+
+    namespace, name = split_iri(DCTERMS.prefLabel)
+    assert namespace == str(DCTERMS)
+    assert name == "prefLabel"
+
+    namespace, name = split_iri(RDFS.subClassOf)
+    assert namespace == str(RDFS)
+    assert name == "subClassOf"
+
+
+def test_parse_literal() -> None:
+    """Test parse n3-encoded literal value."""
+    from tripper import XSD, Literal, parse_literal
+
+    value, lang, datatype = parse_literal(Literal("abc").n3())
+    assert value == "abc"
+    assert lang is None
+    assert datatype == XSD.string
+
+    value, lang, datatype = parse_literal(Literal("abc", lang="en").n3())
+    assert value == "abc"
+    assert lang == "en"
+    assert datatype is None
+
+    value, lang, datatype = parse_literal(Literal(3).n3())
+    assert value == 3
+    assert lang is None
+    assert datatype == XSD.integer
+
+    value, lang, datatype = parse_literal(Literal(3.14).n3())
+    assert value == 3.14
+    assert lang is None
+    assert datatype == XSD.double
+
+    value, lang, datatype = parse_literal(Literal(True).n3())
+    assert value is True
+    assert lang is None
+    assert datatype == XSD.boolean
+
+    value, lang, datatype = parse_literal("3")
+    assert value == 3
+    assert lang is None
+    assert datatype == XSD.integer
+
+    value, lang, datatype = parse_literal("3.14")
+    assert value == 3.14
+    assert lang is None
+    assert datatype == XSD.double
--- a/tests/test_literals.py
+++ b/tests/test_literals.py
@@ -71,39 +71,62 @@ def test_split_iri() -> None:
 
 def test_parse_literal() -> None:
     """Test parse n3-encoded literal value."""
+    from datetime import datetime
+
     from tripper import XSD, Literal, parse_literal
 
-    value, lang, datatype = parse_literal(Literal("abc").n3())
-    assert value == "abc"
-    assert lang is None
-    assert datatype == XSD.string
+    literal = parse_literal(Literal("abc").n3())
+    assert literal.value == "abc"
+    assert literal.lang is None
+    assert literal.datatype == XSD.string
 
-    value, lang, datatype = parse_literal(Literal("abc", lang="en").n3())
-    assert value == "abc"
-    assert lang == "en"
-    assert datatype is None
+    literal = parse_literal(Literal("abc", lang="en").n3())
+    assert literal.value == "abc"
+    assert literal.lang == "en"
+    assert literal.datatype is None
 
-    value, lang, datatype = parse_literal(Literal(3).n3())
-    assert value == 3
-    assert lang is None
-    assert datatype == XSD.integer
+    literal = parse_literal(Literal(3).n3())
+    assert literal.value == 3
+    assert literal.lang is None
+    assert literal.datatype == XSD.integer
 
-    value, lang, datatype = parse_literal(Literal(3.14).n3())
-    assert value == 3.14
-    assert lang is None
-    assert datatype == XSD.double
+    literal = parse_literal(Literal(3.14).n3())
+    assert literal.value == 3.14
+    assert literal.lang is None
+    assert literal.datatype == XSD.double
 
-    value, lang, datatype = parse_literal(Literal(True).n3())
-    assert value is True
-    assert lang is None
-    assert datatype == XSD.boolean
+    literal = parse_literal(Literal(True).n3())
+    assert literal.value is True
+    assert literal.lang is None
+    assert literal.datatype == XSD.boolean
 
-    value, lang, datatype = parse_literal("3")
-    assert value == 3
-    assert lang is None
-    assert datatype == XSD.integer
+    dt = datetime(2022, 10, 23)  # pylint: disable=invalid-name
+    literal = parse_literal(Literal(dt).n3())
+    assert literal.value == dt
+    assert literal.lang is None
+    assert literal.datatype == XSD.dateTime
 
-    value, lang, datatype = parse_literal("3.14")
-    assert value == 3.14
-    assert lang is None
-    assert datatype == XSD.double
+    literal = parse_literal(dt)
+    assert literal.value == dt
+    assert literal.lang is None
+    assert literal.datatype == XSD.dateTime
+
+    literal = parse_literal("abc")
+    assert literal.value == "abc"
+    assert literal.lang is None
+    assert literal.datatype == XSD.string
+
+    literal = parse_literal("3")
+    assert literal.value == 3
+    assert literal.lang is None
+    assert literal.datatype == XSD.integer
+
+    literal = parse_literal("3.14")
+    assert literal.value == 3.14
+    assert literal.lang is None
+    assert literal.datatype == XSD.double
+
+    literal = parse_literal(False)
+    assert literal.value is False
+    assert literal.lang is None
+    assert literal.datatype == XSD.boolean
--- a/tests/test_literals.py
+++ b/tests/test_literals.py
@@ -18,7 +18,7 @@ def test_string() -> None:
 
 def test_string_lang() -> None:
     """Test creating a string literal with a set language."""
-    from tripper import Literal
+    from tripper.literal import Literal
 
     literal = Literal("Hello world!", lang="en")
     assert literal.lang == "en"
--- a/tests/test_literals.py
+++ b/tests/test_literals.py
@@ -1,4 +1,5 @@
 """Test RDF literals."""
+# pylint: disable=invalid-name
 
 
 def test_string() -> None:
@@ -100,7 +101,7 @@ def test_parse_literal() -> None:
     assert literal.lang is None
     assert literal.datatype == XSD.boolean
 
-    dt = datetime(2022, 10, 23)  # pylint: disable=invalid-name
+    dt = datetime(2022, 10, 23)
     literal = parse_literal(Literal(dt).n3())
     assert literal.value == dt
     assert literal.lang is None
--- a/tests/test_literals.py
+++ b/tests/test_literals.py
@@ -29,7 +29,7 @@ def test_string_lang() -> None:
 
 def test_en() -> None:
     """Test creating a string literal through `en()`."""
-    from tripper import en
+    from tripper.utils import en
 
     literal = en("Hello world!")
     assert literal.n3() == '"Hello world!"@en'
@@ -59,7 +59,8 @@ def test_float_through_datatype() -> None:
 
 def test_split_iri() -> None:
     """Test parse n3-encoded literal value."""
-    from tripper import DCTERMS, RDFS, split_iri
+    from tripper import DCTERMS, RDFS
+    from tripper.utils import split_iri
 
     namespace, name = split_iri(DCTERMS.prefLabel)
     assert namespace == str(DCTERMS)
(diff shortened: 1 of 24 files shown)


tests/test_literals.py
```diff
@@ -1,9 +1,10 @@
 """Test RDF literals."""
+# pylint: disable=invalid-name
 
 
 def test_string() -> None:
     """Test creating a string literal."""
-    from tripper.triplestore import Literal
+    from tripper import Literal
```
CasperWA, line 7:
```suggestion
    from tripper.literal import Literal
```
Why use a general import rather than the explicitly from the module we are testing?

This goes for all the changed imports here.

tripper/__init__.py
```diff
@@ -3,7 +3,9 @@
```
CasperWA, line 3: A general question here.
Do we _really_ need to import all of this from the base module? What's the benefit? Is there something that can be left out that isn't truly necessary to expose completely or can be relegated to sub-module imports instead (like, e.g. `en`)?

tripper/backends/ontopy.py
```diff
@@ -10,7 +10,7 @@
 
 from ontopy.ontology import Ontology, _unabbreviate, get_ontology
 
-from tripper.triplestore import Literal
+from tripper import Literal
```
CasperWA, line 13:
```suggestion
from tripper.literal import Literal
```
Again, why not import directly from the module, where it exists, instead of the base package name? If the module is part of the public API, it's the more "accurate" method.

tripper/literal.py
```diff
@@ -0,0 +1,193 @@
+"""Literal rdf values."""
```
CasperWA, line 1: Are there any changes in this content after transferring it here from `triplestore.py`? It might be nice to just to a PR (or a dedicated commit) that has _only_ the split up and move within it, and then a subsequent PR (or dedicated commit) to implement any changes.

tripper/namespace.py
```diff
@@ -0,0 +1,181 @@
+"""Provides a simple representation of namespaces."""
```
CasperWA, line 1: Are there any changes in this content after transferring it here from `triplestore.py`? It might be nice to just to a PR (or a dedicated commit) that has _only_ the split up and move within it, and then a subsequent PR (or dedicated commit) to implement any changes.

tripper/triplestore.py
```diff
@@ -18,10 +18,13 @@
 import re
 import warnings
 from collections.abc import Sequence
-from datetime import datetime
 from importlib import import_module
 from typing import TYPE_CHECKING
 
+from .errors import NamespaceError, TriplestoreError, UniquenessError
+from .literal import Literal, en
+from .namespace import DCTERMS, DM, FNO, MAP, OWL, RDF, RDFS, XML, XSD, Namespace
```
CasperWA, line 26:
```suggestion
from tripper.errors import NamespaceError, TriplestoreError, UniquenessError
from tripper.literal import Literal, en
from tripper.namespace import DCTERMS, DM, FNO, MAP, OWL, RDF, RDFS, XML, XSD, Namespace
```
Relative imports should not be used if it can be helped. The way they work has been changed throughout the minor Python 3 versions and it is not as runtime "stable" as absolute imports.
`__init__.py` files are the exception to this rule.

tripper/triplestore.py
```diff
@@ -688,11 +381,10 @@ def add_function(
 
         method = getattr(self, f"_add_function_{standard}")
         func_iri = method(func, expects, returns, base_iri)
-        self.function_repo[func_iri] = func
-
+        if callable(func):
+            self.function_repo[func_iri] = func
```
CasperWA, line 385: Before, this was always set - is there a point in the usage of an instance of this class that this fact was expected?

tripper/triplestore.py
```diff
@@ -763,6 +470,21 @@ def _add_function_fno(self, func, expects, returns, base_iri):
...
+def function_id(func: "Callable", length: int = 4) -> str:
+    """Return a checksum for function `func`.
+
+    The returned object is a string of hexadecimal digits.
+
+    `length` is the number of bytes in the returned checksum.  Since
+    the current implementation is based on the shake_128 algorithm,
+    it make no sense to set `length` larger than 32 bytes.
+    """
+    # return hex(crc32(inspect.getsource(func).encode())).lstrip('0x')
```
CasperWA, line 482: Can this line be removed?

tripper/triplestore.py
```diff
@@ -784,16 +506,14 @@ def infer_iri(obj):
     raise TypeError("cannot infer IRI from object {obj!r}")
 
 
-def function_id(func, length=4):
-    """Return a checksum for function `func`.
+def split_iri(iri: str) -> "Tuple[str, str]":
+    """Split iri into namespace and name parts and return them as a tuple."""
```
CasperWA, line 510:
```suggestion
    """Split iri into namespace and name parts and return them as a tuple.

    Parameters:
        iri: The IRI to be split.

    Returns:
        A split IRI. Split into namespace and name.

    """
```

tripper/triplestore.py
```diff
@@ -784,16 +506,14 @@ def infer_iri(obj):
...
 
-    `length` is the number of bytes in the returned checksum.  Since
-    the current implementation is based on the shake_128 algorithm,
-    it make no sense to set `length` larger than 32 bytes.
-    """
-    # return hex(crc32(inspect.getsource(func).encode())).lstrip('0x')
-    return hashlib.shake_128(  # pylint: disable=too-many-function-args
-        inspect.getsource(func).encode()
-    ).hexdigest(length)
+    raise ValueError("all IRIs should contain a slash")
```
CasperWA, line 519:
```suggestion
    raise ValueError("all IRIs should contain a forward slash or a hash")
```

CasperWA (changes requested):

tests/test_literals.py
jesper-friis: Lets address this when we have agreed on the next question

tripper/__init__.py
jesper-friis: Agree, maybe I have included too much. But I think that the namespaces and the three core classes Triplestore, Literal and Namespace should at least be in the __init__.py file. 

The errors are are well located in errors.py
If the utility functions, en(), parse_literal() and split_iri() should not be imported by __init__.py, I think they should be moved into a new module called utils.py.

tripper/literal.py
jesper-friis: Yes, some bug fixing and adding parse_literal(). The PR could arguably have been cleaner if it was split up. I just ended up the possibility to do the refactoring while I anyway was messing around with tripper

tripper/namespace.py
jesper-friis: Maybe some bug fixes

tripper/triplestore.py
jesper-friis: Yes. In OntoFlow we want to semantically describe a model or conversion function without necessarily providing an implementation of it. To support that, `func` can either be a callable or just an IRI that stands for the model or function.

tripper/triplestore.py
jesper-friis: done

tripper/triplestore.py
jesper-friis: Is that really true? "a#b" is not a valid IRI (even though that the current implementation will happily split it...).

tripper/triplestore.py
jesper-friis: I committed this change, but I thought that the Google style docstrings used "Arguments:" instead of "Parameters:". 

However, when I checked it up, it turns up to be "Args:". See
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

They also places "Note:" above the arguments - which seems odd and is the opposite of the NumPy docstrings...

tripper/__init__.py
CasperWA: > (...) If the utility functions, en(), parse_literal() and split_iri() should not be imported by **init**.py, I think they should be moved into a new module called utils.py.

This is a great idea. They should probably have been put into such a file even if the goal _wasn't_ to get a dedicated sub-module for imports, but just to separate concerns and clean up the `triplestore.py` file further.

tripper/literal.py
CasperWA: It's quite difficult/extra time consuming for me (or anyone) to review such a mixed change, since it requires one to go through the entire file from top to bottom as if it was a completely newly written file. The separation would have eased this, focusing in on only the real changes.

tripper/namespace.py
CasperWA: Which?

tripper/triplestore.py
CasperWA: Sorry, my question was not clear. I'll try again.
Prior to this change, the value `self.function_repo[func_iri]` was _always_ set. However, now it is _only_ set if `func` is a callable.
Is there any point in the code base, where this value might be expected to be present without actually performing a check for it? If yes, this value should probably be instantiated with `None` or set to `None` in an `else` block, in order to avoid a `KeyError` (and instead raise a more understandable `ValueError` or similar?).

tripper/triplestore.py
CasperWA: Indeed you're right:

https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

tripper/triplestore.py
CasperWA: I was merely updating the error message according to the logic you were implementing in the function here. You were first checking for a hash, and then a forward slash, returning out of the function after each check, and so I put this logic into the error message for a bit of transparency for the developer/user who eventually reads it and needs to understand why something hasn't worked :)

tripper/literal.py
```diff
@@ -0,0 +1,193 @@
...
+            except TypeError:
+                pass
+        return Literal(value, lang=lang, datatype=datatype)
+
+    for type_, datatype in Literal.datatypes.items():
+        if type_ is not bool:
+            try:
+                return Literal(type_(literal), lang=lang, datatype=datatype)
+            except (ValueError, TypeError):
+                pass
```
CasperWA, line 191: Why do we allow these exceptions to go uncaught? Wouldn't that constitute not being able to parse the `literal` and we should therefore raise?

tripper/namespace.py
```diff
@@ -0,0 +1,181 @@
...
+from .errors import NamespaceError, NoSuchIRIError
```
CasperWA, line 2:
```suggestion
from tripper.errors import NamespaceError, NoSuchIRIError
```

tripper/namespace.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def __init__(
+        self,
+        iri,
+        label_annotations=(),
+        check=False,
+        cachemode=-1,
+        triplestore=None,
+        triplestore_url=None,
+    ):
```
CasperWA, line 53:
```suggestion
        iri: str,
        label_annotations: "Optional[Sequence[str]]" = None,
        check: bool = False,
        cachemode: int = -1,
        triplestore: "Optional[Triplestore]" = None,
        triplestore_url: "Optional[str]" = None,
    ) -> None:
```

tripper/namespace.py
```diff
@@ -0,0 +1,181 @@
...
+        if label_annotations is True:
```
CasperWA, line 54:
```suggestion
        if label_annotations is not None:
```

tripper/namespace.py
```diff
@@ -0,0 +1,181 @@
...
+        self._label_annotations = tuple(label_annotations)
+        self._check = bool(check)
+
+        need_triplestore = bool(check or label_annotations)
+        if cachemode == -1:
+            cachemode = Namespace.ONLY_CACHE if need_triplestore else Namespace.NO_CACHE
+
+        if need_triplestore and triplestore is None:
+            # Import Triplestore here to break cyclic-import
+            from .triplestore import (  # pylint: disable=import-outside-toplevel,cyclic-import
```
CasperWA, line 67:
```suggestion
            from tripper.triplestore import (  # pylint: disable=import-outside-toplevel
```

tripper/namespace.py
```diff
@@ -0,0 +1,181 @@
...
+            for s in triplestore.subjects(RDFS.isDefinedBy, self._iri)
+            if s.startswith(self._iri)
+        )
+
+    def __getattr__(self, name):
+        if self._cache and name in self._cache:
+            return self._cache[name]
+
+        if self._triplestore:
+
```
CasperWA, line 122:
```suggestion
```

tripper/namespace.py
```diff
@@ -0,0 +1,181 @@
...
+            # Check if ``iri = self._iri + name`` is in the triplestore.
+            # If so, add it to the cache.
+            # We only need to check that generator returned by
```
CasperWA, line 125:
```suggestion
            # We only need to check that the generator returned by
```

CasperWA (changes requested):

tripper/__init__.py
jesper-friis: done

tripper/triplestore.py
jesper-friis: Not in tripper or OntoFlow. But if you try to instantiate a datamodel that via a missing mapping function DLite will fail...

Hmm, setting it to None is actually a good idea. Then DLite would be able to come with a better error message. 
Implemented!

tripper/triplestore.py
jesper-friis: An IRI does not need a hash, although most ontologies uses a hash to separate the namespace from the name part.
A prominent exception is dcterms, which uses a slash...

tripper/triplestore.py
```diff
@@ -724,24 +416,39 @@ def _add_function_fno(self, func, expects, returns, base_iri):
...
+            func_iri = f"{base_iri}{func.__name__}_{fid}"
+            doc_string = inspect.getdoc(func)
+            parlist = f"_:{func.__name__}{fid}_parlist"
+            outlist = f"_:{func.__name__}{fid}_outlist"
+            if isinstance(expects, Sequence):
+                pars = list(zip(expects, inspect.signature(func).parameters))
+            else:
+                pars = [
+                    (expects[par], par) for par in inspect.signature(func).parameters
+                ]
```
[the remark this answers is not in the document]
jesper-friis: Yes, I like to shorten - short line lengths greatly improves readability for me.
Furthermore fits the use of `par` well with other variable names in the same method: `parlist`, `pariris`, `parnames`...

tripper/triplestore.py
```diff
@@ -724,24 +416,39 @@ def _add_function_fno(self, func, expects, returns, base_iri):
...
+        elif isinstance(func, str):
+            func_iri = func
+            doc_string = ""
+            parlist = f"_:{func_iri}_parlist"
+            outlist = f"_:{func_iri}_outlist"
+            pariris = expects if isinstance(expects, Sequence) else expects.values()
+            parnames = [split_iri(par)[1] for par in pariris]
```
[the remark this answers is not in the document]
jesper-friis: I agree, in this case we should use `pariri` instead of `par`.

tripper/literal.py
```diff
@@ -0,0 +1,193 @@
...
+        datetime: XSD.dateTime,
+        bytes: XSD.hexBinary,
+        bytearray: XSD.hexBinary,
+        bool: XSD.boolean,
+        int: XSD.integer,
+        float: XSD.double,
+        str: XSD.string,
+    }
+
+    def __new__(cls, value, lang=None, datatype=None):
```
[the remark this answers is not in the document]
jesper-friis: > Unsure whether the return type here should actually be `"Literal"` instead?

Yes, it should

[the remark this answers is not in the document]
jesper-friis: Mypy is just too stupid to for this. It correctly requires that __new__() should return Literal. But then it complains about line 58 where we assign a string to the return value (which is fully ok since Literal is a subclass of str). Unless there is a way to locally shut up mypy, __new__() should be left without annotations.

tripper/literal.py
```diff
@@ -0,0 +1,193 @@
...
+            return f'"{self}"@{self.lang}'
+        if self.datatype:
+            return f'"{self}"^^{self.datatype}'
+        return f'"{self}"'
+
+
+def en(value):  # pylint: disable=invalid-name
+    """Convenience function that returns value as a plain english literal.
+
+    Equivalent to``Literal(value, lang="en")``.
```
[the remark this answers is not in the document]
jesper-friis: I learnt that single backtick indicates a variable, while double backtick indicate a code literal.

tripper/literal.py
```diff
@@ -0,0 +1,193 @@
...
+    return Literal(value, lang="en")
+
+
+def parse_literal(literal: "Any") -> "Literal":
+    """Parse `literal` and return it as an instance of Literal.
+
+    The main difference between this function and the Literal constructor,
+    is that this function correctly interprets n3-encoded literal strings.
+    """
+    # pylint: disable=invalid-name,too-many-branches
```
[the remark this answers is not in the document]
jesper-friis: There is no semantics in the return value of match.groups() - hence generic the variable names `v1` and `v2` are completely fine. Making them longer reduces readability for no gain.

tripper/literal.py
```diff
@@ -0,0 +1,193 @@
...
+            )
+            if match:
+                _, v1, v2, lang = match.groups()
+                value = v1 if v1 else v2
+            else:
+                value = literal
+
+    if lang or datatype:
+        if datatype:
+            types = {v: k for k, v in Literal.datatypes.items()}
```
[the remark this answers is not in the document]
jesper-friis: For me this reduces readability.

tripper/literal.py
```diff
@@ -0,0 +1,193 @@
...
+            type_ = types[datatype]
+            try:
+                value = type_(value)
```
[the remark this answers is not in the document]
jesper-friis: No. We wan't to minimise what we put in the try clause. We wan't to accept a `TypeError` from `type_(value)`, but not from `types[datatype]`.

tripper/literal.py
jesper-friis: The literal does not provide datatype, so we try to infer the datatype by looping over all supported datatypes. If we get a ValueError or TypeError it is not the correct datatype and we try again...

tripper/namespace.py
jesper-friis: When adding commit suggestions, please make sure that they work. If you are not sure, write out your suggestion in text.

tripper/namespace.py
jesper-friis: The value `True` is explicitly stated in the docstring.

tripper/namespace.py
jesper-friis: Have you tested this suggestion? It makes pylint fail...

PR #29, opened 2022-11-02
Moved CI/CD workflows depending on SINTEF to another folder

jesper-friis:
Incompatibilities between the CI/CD tools breaks all commits. To work around this we turn off the CI/CD using the new SINTEF repo for now.

The real issue is that mkdocs requires all methods to be annotated.

However, when adding the required annotations, the `__new__()`method should return an instance of the class. But mypy don't understand that it is completely valid for a `__new__()` method of a class that subclasses str to assign the return value to a str.

This is a bug in mypy, but it blocks any merge of PRs.


jesper-friis:
Fixed in PR #32 

Closing

PR #39, opened 2022-11-20
Added create_database() and remove_database() methods.

### LINKED ISSUE #34: Implement create_database() and remove_database() methods.
They should be optional for the backend to implement, depending on whether they support the concept of several databases.

Connection to a database is done in the `__init__()` method.

jesper-friis:
Closes #34

### DIFF
--- a/tripper/backends/ontopy.py
+++ b/tripper/backends/ontopy.py
@@ -27,6 +27,7 @@ class OntopyStrategy:
     Arguments:
         base_iri: The base iri of the ontology.
             Default to "http://example.com/onto#" if `onto` is not given.
+        database: Unused - ontopy does not support multiple databases.
         onto: Ontology to initiate the triplestore from.  Defaults to an new
             ontology with the given `base_iri`.
         load: Whether to load the ontology.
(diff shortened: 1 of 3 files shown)


CasperWA (commented):
I think we need to come up with a better way of updating/changing the initialization interface for the backends. Adding an unused parameter to all of them whenever one changes seems extremely inefficient.

xAlessandroC (commented):
To me, the changes are fine. Since I still have to start the integration with OntoREC, do you need more proper code testing for the final merge of these features?

I also share Casper's point, maybe this approach won't scale in the future.

jesper-friis:
> To me, the changes are fine. Since I still have to start the integration with OntoREC, do you need more proper code testing for the final merge of these features?
> 
> I also share Casper's point, maybe this approach won't scale in the future.

OK, I removed the explicit explicit database argument from Triplestore.__init__()

CasperWA:
> > To me, the changes are fine. Since I still have to start the integration with OntoREC, do you need more proper code testing for the final merge of these features?
> > I also share Casper's point, maybe this approach won't scale in the future.
> 
> OK, I removed the explicit explicit database argument from Triplestore.**init**()

I don't think this was the point. The point was to design a replacement interface and procedure of handling this. This issue is not one that is to be solved in _this_ PR, but rather a new issue should be opened and this interface and procedure should be updated in another PR.

jesper-friis:
> I don't think this was the point. The point was to design a replacement interface and procedure of handling this. This issue is not one that is to be solved in _this_ PR, but rather a new issue should be opened and this interface and procedure should be updated in another PR.

This PR provides a common API for backends with multiple databases. It does not require anything change in code in backends that doesn't support multiple databases.

CasperWA (approved):

PR #40, opened 2022-11-21
Updated import statements in the tutorial

jesper-friis:
Simple PR correcting the tutorial.

Maybe we could consider to make a test that run doctest on it to ensure that it actually works.

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -51,7 +51,7 @@ New triples can be added either with the `parse()` method (for backends that sup
 
 ```python
 # en(msg) is a convenient function for adding english literals.
-# It is equivalent to ``triplestore.Literal(msg, lang="en")``.
+# It is equivalent to ``tripper.Literal(msg, lang="en")``.
 from tripper.utils import en
 ts.parse("onto.ttl", format="turtle")
 ts.add_triples([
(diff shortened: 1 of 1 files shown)


CasperWA (approved):
Does `doctest` run the Python code? What is `doctest`?

docs/tutorial.md
```diff
@@ -52,7 +52,7 @@ New triples can be added either with the `parse()` method (for backends that sup
 ```python
 # en(msg) is a convenient function for adding english literals.
 # It is equivalent to ``triplestore.Literal(msg, lang="en")``.
```
CasperWA, line 54:
```suggestion
# It is equivalent to ``tripper.triplestore.Literal(msg, lang="en")``.
```

CasperWA (changes requested):

jesper-friis:
> Does `doctest` run the Python code? What is `doctest`?

[doctest](https://docs.python.org/3/library/doctest.html) runs Python code in docstrings. Good for ensuring that the documentation is up-to-date and that examples actually works.

CasperWA:
> > Does `doctest` run the Python code? What is `doctest`?
> 
> [doctest](https://docs.python.org/3/library/doctest.html) runs Python code in docstrings. Good for ensuring that the documentation is up-to-date and that examples actually works.

It seems to only work for Python shell doc strings (starting with `>>>`). Perhaps this one will be better: https://github.com/tmarktaylor/phmdoctest ?

CasperWA (approved):

CasperWA:
Squash+merge at will.

PR #44, opened 2022-12-01
Added collection backend

### LINKED ISSUE #41: Add DLite collection backend
This is easily implemented and would be the right way for mapping and function/transformation strategies to store and utilise the mappings in a pipeline. 

Only the  triples(), add_triples() and remove() methods needs to be implemented.

jesper-friis:
Closes #41 

Simple backend providing a tripper interface to DLite collections. Will be used by oteapi when DLite is chooses as interoperability platform.

This PR also include tests for utils.py - they are written as a script since they are so simple that test isolation via pytest is not needed. 

Note this is a minimal implementation. Support for prefixes should be added in follow-up PRs.

Right now all triples are stored written out completely, but for the user of dlite, I guess it would be easier if the IRIs stored in the collection would be prefixed. This is a topic for a later PR, though.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -32,6 +32,7 @@ with pytest.raises(TypeError):
 with pytest.raises(ValueError):
     split_iri("abc")
 
+
 # Test function_id()
 def f():
     """Function."""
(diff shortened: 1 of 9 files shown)


tests/test_utils.py
```diff
@@ -0,0 +1,86 @@
+"""Test utils"""
+# pylint: disable=invalid-name
```
quaat, line 2: not necessary. remove

tripper/utils.py
```diff
@@ -133,3 +140,41 @@ def parse_literal(literal: "Any") -> "Literal":
...
+        if re.match(r"^[a-z]+://", object):
+            return object
+        types = {
+            int: XSD.integer,
+            float: XSD.double,
+            datetime.datetime.fromisoformat: XSD.dateTime,
+        }
+        for pytype, datatype in types.items():
+            try:
+                pytype(object)  # type: ignore
```
quaat, line 174: will never reach 'float'.

quaat (changes requested):
Clean up code, use regexp for parse_object. Try to avoid redefining built-ins

tests/test_utils.py
jesper-friis: Unfortunately, pylint is really picky and don't understand that this is a script. It complains for instance about variables not being purely upper case. Some of the tools have their own non-standard, non-PEP8 notion of best practices that I do not always agree with. The non-80 character line length (that is used by all large python projects and is described in PEP8 and the google style guidelines) is one example. Another is this requirement that all variable names in a test script must be purely upper case.

tripper/utils.py
jesper-friis: It actually does. It is tested for in tests/test_units.py

quaat: Fair enough, but then I would rather prefer some construct like this, to separate out the nasty `except: pass` in a more readable function and then explicitly test the object for attributes.

```
def isnumber(num):
    try:
        float(num)
        return True
    except ValueError:
        return False
    
...
    if isnumber(str):
        if str.isdigit():
            datatype=XSD.integer
        else:
            datatype = XSD.double
    else:
        ...
        datatype = XSD.datetime
       
```

PR #46, opened 2022-12-03
Fix utils.parse_object()

### LINKED ISSUE #45: Fix utils.parse_object()
`utils.parse_object()` fails for some special cases as noted in the comment at the bottom of `tests/test_utils.py`. Fix `parse_object()` such that it also works for these cases.

Also, as noted in the review of PR #44: clean up code, use regexp for parse_object. Try to avoid redefining built-ins.

jesper-friis:
Closes #45 

General cleanup
- Fixing failing tests by improved generality of parse_literal() and parse_object().
- Added `__eq__()` and `__hash__()` methods to Literal. The `__eq__()` method is required for proper testing...
- Reduced code duplication.
- Hopefully also addressed the comments by @quaat  in PR #44

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -112,7 +112,5 @@ assert parse_object(str(XSD)) == str(XSD)
 assert parse_object(XSD.int) == XSD.int
 assert parse_object(f'"42"^^{XSD.integer}') == Literal("42", datatype=XSD.integer)
 assert parse_object(f'"4.2"^^{XSD.double}') == Literal("4.2", datatype=XSD.double)
-
-# __FIXME__: parse_object() currently fails for the following cases:
-# assert parse_object(f'"42"^^{XSD.double}') == Literal("42", datatype=XSD.double)
-# assert parse_object(f'"42"^^{XSD.int}') == Literal("42", datatype=XSD.int)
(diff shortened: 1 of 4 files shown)


tests/test_utils.py
```diff
@@ -20,6 +20,47 @@
 assert infer_iri(coll.meta) == coll.meta.uri
 assert infer_iri(coll) == coll.uuid
 
+# We have no dependencies on pydantic, hence don't assume that it is installed
```
quaat, line 23: A good starting point could be to use (or expand on) this datamodel: https://github.com/SINTEF/soft7/blob/main/s7/pydantic_models/soft7.py

quaat (dismissed):

tests/test_utils.py
jesper-friis: OK, I renamed `uri` to `identity` in the test. With that it seems that we are pretty much matching the soft7 datamodel. 

However, it is confusing and not very useful to have two different names for the same thing in soft7 and dlite. I have nothing in principle against updating dlite, but it would take several hours to change `uri` to `identity` allover dlite when also keeping backward compatibility. It is time I don't have. Great if someone would like to do it, otherwise I would suggest to update soft7 instead.

quaat (approved):

PR #49, opened 2022-12-11
Separated `base_iri` argument from `triplestore_url` in rdflib backend

jesper-friis:
Before gave rdflib the `base_iri` argument an additional meaning not documented in Triplestore.__init__(). This is now taken over by the `triplestore_url` argument.

### DIFF
--- a/tripper/backends/rdflib.py
+++ b/tripper/backends/rdflib.py
@@ -42,12 +42,11 @@ class RdflibStrategy:
     """Triplestore strategy for rdflib.
 
     Arguments:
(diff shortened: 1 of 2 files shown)


CasperWA (dismissed):

daniel-sintef (dismissed):
I don't think backends should silently allow parameters they don't use. As this could confuse users (imagine one trying to figure out why the base_iri doesn't work, because they need to use triplestore_url). 

I would suggest including a warning if the base_iri is provided, that it is unused.

PR #51, opened 2022-12-11
Simplify use of the Triplestore triples() and remove() methods

### LINKED ISSUE #50: Simplify use of the Triplestore triples() and remove() methods
The Triplestore triples() and remove() methods were originally modelled after the corresponding methods in rdflib. However, they are unnecessarily inconvenient to use. 

By changing the prototype from `triples(self, triple)` to `triples(self, subject=None, predicate=None, object=None)` would make the API more intuitive and easier to use. For instance, looping over all instances would simply be `triples()` instead of `triples((None, None, None))`.  The same apply for remove().

The backend interface should be unchanged.

To ensure backward compatibility, we should allow `subject` to be either a string or a triple. We could also add an additional `triple` keyword argument. Providing a triple in `subject` or using the `triple` keyword argument should raise a deprecation warning.

jesper-friis:
Closes #50 

Changed arguments of the Triplestore triples() and remove() methods to make them easier to use.

Ensured backward compatibility.

### DIFF
--- a/tripper/triplestore.py
+++ b/tripper/triplestore.py
@@ -91,15 +91,40 @@ class Triplestore:
 
     # Methods implemented by backend
     # ------------------------------
-    def triples(self, triple: "Triple") -> "Generator[Triple, None, None]":
+    def triples(  # pylint: disable=redefined-builtin
+        self,
+        subject: "Optional[Union[str, Triple]]" = None,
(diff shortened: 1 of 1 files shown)


daniel-sintef (approved):
Create an issue to create something remove code duplication.

PR #62, opened 2023-01-09
Added mappings

jesper-friis:
Added mappings to tripper.

### DIFF
--- a/tests/mappings/test_mappings.py
+++ b/tests/mappings/test_mappings.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+import sys
+from pathlib import Path
+
+from tripper import Triplestore
+from tripper.mappings import mapping_routes
+
+# Configure paths
+thisdir = Path(__file__).parent.absolute()
+
+ts = Triplestore(backend="rdflib")
+
+
+# Define some prefixed namespaces
+CHEM = ts.bind("chem", "http://onto-ns.com/onto/chemistry#")
+MOL = ts.bind("mol", "http://onto-ns.com/meta/0.1/Molecule#")
+MOL2 = ts.bind("mol2", "http://onto-ns.com/meta/0.1/Molecule2#")
+SUB = ts.bind("sub", "http://onto-ns.com/meta/0.1/Substance#")
+
+
+# Add mappings from data models to ontology
+ts.add_mapsTo(CHEM.Identifier, MOL.name)
+ts.add_mapsTo(CHEM.Identifier, SUB.id)
+
+ts.add_mapsTo(CHEM.GroundStateEnergy, MOL.groundstate_energy)
+ts.add_mapsTo(CHEM.GroundStateEnergy, MOL2.energy)
+ts.add_mapsTo(CHEM.GroundStateEnergy, SUB.molecule_energy)
+
+
+routes = mapping_routes(
+    target=SUB.molecule_energy,
+    sources={},  # MOL.groundstate_energy: 1.0},
+    triplestore=ts,
+)
--- a/tripper/mappings/mappings.py
+++ b/tripper/mappings/mappings.py
@@ -1,0 +1,744 @@
+"""Implements mappings between entities.
+
+Units are currently handled with pint.Quantity.  The benefit of this
+compared to explicit unit conversions, is that units will be handled
+transparently by mapping functions, without any need to specify units
+of input and output parameters.
+
+Shapes are automatically handled by expressing non-scalar quantities
+with numpy.
+
+"""
+from __future__ import annotations
+
+import itertools
+import types
+import warnings
+from collections import defaultdict
+from collections.abc import Sequence
+from enum import Enum
+from typing import Any, Callable, Dict, List, Optional, Union
+
+import numpy as np
+from pint import Quantity  # remove
+
+from tripper import DM, FNO, MAP, RDF, RDFS, Triplestore
+
+# from dlite.utils import infer_dimensions
+
+
+class MappingError(Exception):
+    """Base class for mapping errors."""
+
+
+class InsufficientMappingError(MappingError):
+    """There are properties or dimensions that are not mapped."""
+
+
+class AmbiguousMappingError(MappingError):
+    """A property maps to more than one value."""
+
+
+class InconsistentDimensionError(MappingError):
+    """The size of a dimension is assigned to more than one value."""
+
+
+class InconsistentTriplesError(MappingError):
+    """Inconsistcy in RDF triples."""
+
+
+class MissingRelationError(MappingError):
+    """There are missing relations in RDF triples."""
+
+
+class StepType(Enum):
+    """Type of mapping step when going from the output to the inputs."""
+
+    MAPSTO = 1
+    INV_MAPSTO = -1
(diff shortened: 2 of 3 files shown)


tests/mappings/test_mappings.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python
+"""Test mappings."""
```
francescalb, line 2: Have we not agreed on using pytest for making tests?

tests/mappings/test_mappings.py
```diff
@@ -0,0 +1,34 @@
...
+ts.add_mapsTo(CHEM.GroundStateEnergy, MOL.groundstate_energy)
+ts.add_mapsTo(CHEM.GroundStateEnergy, MOL2.energy)
+ts.add_mapsTo(CHEM.GroundStateEnergy, SUB.molecule_energy)
+
+
+routes = mapping_routes(
+    target=SUB.molecule_energy,
+    sources={},  # MOL.groundstate_energy: 1.0},
+    triplestore=ts,
+)
```
francescalb, line 34: I suggest adding some asserts at the end here to check that the mapping_routes that are returned are the ones we expect. 

Also, all the different options should be tested, e.g. changing cost or getting the number of routes.

tripper/mappings/mappings.py
```diff
@@ -0,0 +1,770 @@
...
+        step = step0
+        target = source
+    if target not in soMaps:
+        raise MissingRelationError(f'Missing "mapsTo" relation on: {target}')
+    walk(target, visited, step)
+
+    return step
+
+
+# def instance_routes(
```
francescalb, line 621: I suggest adding a comment for why this big chunk of code is commented out. And maybe also write an issue about it.

tests/mappings/test_mappings.py
jesper-friis: I think we should use pytest when it make sense to do so. For simple cases pytest just make the tests more difficult to understand and makes it impossible to use the tests for interactive development and debugging. The latter greatly improves my development efficiency of good code. 

When running the test scripts from pytest, we still get most of the advantages of pytest, autodetection of tests, running several tests in parallell, skipping selected tests, etc...

However, I still think pytest may be the right tool for complex cases where test separation is important (even though that can be achieved by splitting the tests into several files) when we need parametrised tests, or similar. 

But in this case, I think a test script is superior.

tests/mappings/test_mappings.py
jesper-friis: done

tripper/mappings/mappings.py
jesper-friis: The commented out code depends on dlite and should therefore be kept in dlite. I have now removed it.

tripper/mappings/mappings.py
```diff
@@ -0,0 +1,770 @@
...
+    ):  # pylint: disable=unused-argument
+        """Returns a string representation of the Value.
+
+        Arguments:
+            routeno: Unused.  The argument exists for consistency with
+                the corresponding method in Step.
+            name: Name of value.
+            indent: Indentation level.
+        """
+        s = []
```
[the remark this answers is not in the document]
jesper-friis: It is a temporary list of strings that we progressively build up before we finally join them for return. 

I am not sure what would be a good name. Maybe `strings`, but I am not sure whether that is very helpful either. Do you think so?

tripper/mappings/mappings.py
```diff
@@ -0,0 +1,618 @@
...
+
+    # pylint: disable=too-many-arguments,too-many-locals,too-many-statements
+
+    if function_repo is None:
+        function_repo = triplestore.function_repo
+
+    default_costs = dict(default_costs)
+
+    # Create lookup tables for fast access to properties
+    # This only transverse `tiples` once
```
francescalb, line 519:
```suggestion
    # This only traverses `triples` once
```

tripper/mappings/mappings.py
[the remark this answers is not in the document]
francescalb: Yes, I think it is better. Then it is clear from the name what it is supposed to contain.

tripper/mappings/mappings.py
```diff
@@ -0,0 +1,618 @@
...
+                    )
+                    step.add_input(step0, name=soName.get(input_iri))
+                    walk(input_iri, visited, step0)
+                step.join_input()
+
+    visited = set()
+    step = MappingStep(output_iri=target, output_unit=soUnit.get(target))
+    if target in soInst:
+        # It is only initially we want to follow instanceOf in forward
+        # direction.
```
francescalb, line 605: Why? I do not understand this comment.

tripper/mappings/mappings.py
```diff
@@ -0,0 +1,618 @@
...
+        visited.add(target)  # do we really wan't this?
```
francescalb, line 606: Why or why not? I think this comment is confusing.

.pre-commit-config.yaml
```diff
@@ -33,7 +33,7 @@ repos:
     rev: 1.7.4
     hooks:
     - id: bandit
-      args: ["-r"]
+      args: ["--skip=B101", "-r"]
```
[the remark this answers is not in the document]
francescalb: Ok, but we have to remember that this is code-base with a lot of developers with varying level of expertise. Could we be more specific on which files this bandit test is skipped?

"Know what we are doing" is important. We should have a section on this in the demo.

tripper/mappings/mappings.py
jesper-friis: Good question. It may actually be that we not only want to follow instanceOf initially. I think we need to make some drawings...

tests/mappings/test_mappings.py
francescalb: Ok, we really need to decide on a team level what we go for, but I will approve this for now.

tripper/mappings/mappings.py
```diff
@@ -0,0 +1,640 @@
...
+        self.function = function
+        self.cost = cost
+        self.output_unit = output_unit
+        self.input_routes: "List[dict]" = []  # list of inputs dicts
+        self.join_mode = False  # whether to join upcoming input
+        self.joined_input: "Inputs" = {}
+
+    def add_inputs(self, inputs: "Inputs"):
+        """Add input dict for an input route."""
+        assert isinstance(inputs, dict)  # nosec
```
francescalb, line 164:
```suggestion
        assert isinstance(inputs, dict)  # nosec B101
```
To specify what test we are skipping.

tripper/mappings/mappings.py
```diff
@@ -0,0 +1,640 @@
...
+        """Add an input (MappingStep or Value), where `name` is the name
+        assigned to the argument.
+
+        If the `join_mode` attribute is false, a new route is created with
+        only one input.
+
+        If the `join_mode` attribute is true, the input is remembered, but
+        first added when join_input() is called.
+        """
+        assert isinstance(input, (MappingStep, Value))  # nosec
```
francescalb, line 177:
```suggestion
        assert isinstance(input, (MappingStep, Value))  # nosec B101
```

PR #67, opened 2023-01-25
Updated Triplestore.add_function() to also support EMMO.

### LINKED ISSUE #65: Describe functions with EMMO instead of FnO
Add an _add_function_emmo() to triplestore.py and emmo_mapper() to mappings/mappings.py.
Make EMMO the default.

jesper-friis:
Closes #65


tripper/triplestore.py
```diff
@@ -646,3 +658,55 @@ def _add_function_fno(self, func, expects, returns, base_iri):
...
+        self.bind("dcterms", DCTERMS)
+        self.bind("map", MAP)
+
+        # Hardcode EMMO IRIs to avoid slow lookup
+        Task = EMMO.EMMO_4299e344_a321_4ef2_a744_bacfcce80afc
+        DataSet = EMMO.EMMO_194e367c_9783_4bf5_96d0_9ad597d48d9a
+        hasInput = EMMO.EMMO_36e69413_8c59_4799_946c_10b05d266e22
+        hasOutput = EMMO.EMMO_c4bace1d_4db0_4cd3_87e9_18122bae2840
+        # Software = EMMO.EMMO_8681074a_e225_4e38_b586_e85b0f43ce38
+        # hasSoftware = EMMO.Software  # TODO: fix
```
daniel-sintef, line 675: How hard would this be to fix? Can you provide more information?

tripper/triplestore.py
```diff
@@ -646,3 +658,55 @@ def _add_function_fno(self, func, expects, returns, base_iri):
             lst = lst_next
 
         return func_iri
+
+    def _add_function_emmo(self, func, expects, returns, base_iri):
+        """Implementing add_function() for EMMO."""
```
daniel-sintef, line 663: Do we need a docstring?

tests/test_add_function.py
```diff
@@ -0,0 +1,119 @@
+"""Test Triplestore.add_function()"""
+# pylint: disable=invalid-name
```
daniel-sintef, line 2: Very glad to see test added with the code :)

tripper/triplestore.py
jesper-friis: It requires that we add the hasSoftware object property to EMMO - easy to do, but will take time since it should be quickly discussed in the EMMO team

daniel-sintef: No problem. Just write something like "requires the hasSoftware object property which is not currently implemented in EMMO". This will make it easier if someone other than you ends up fixing the todo :)

tripper/triplestore.py
jesper-friis: This is an EMMO-specific implementation of the public method `add_function()`, which is documented above. 

I will add a few more words...

PR #68, opened 2023-01-25
Proper cost function

jesper-friis:
This PR builds on top of PR #67 

It makes the cost function independent of any data, such that the costs can be calculated purely from the triplestore (from which data may be accessed). This is needed by OntoFlow.

### DIFF
--- a/tripper/mappings/mappings.py
+++ b/tripper/mappings/mappings.py
@@ -327,9 +327,6 @@ class MappingStep:
                     input_iris = [input.output_iri for input in inputs.values()]
                     owncost = self.cost(self.triplestore, input_iris, self.output_iri)
                     base.cost[i] += owncost
-                    # values = get_values(inputs, rno, magnitudes=True)
-                    # owncost = self.cost(**values)
-                    # base.cost[i] += owncost
             else:
                 owncost = self.cost
                 base.cost += owncost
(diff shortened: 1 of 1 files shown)


tripper/mappings/mappings.py
```diff
@@ -318,9 +323,13 @@ def lowest_costs(self, nresults: int = 5) -> "List[Tuple[float, int]]":
             # mapping step.
             if callable(self.cost):
                 for i, rno in enumerate(base.routeno):
-                    values = get_values(inputs, rno, magnitudes=True)
-                    owncost = self.cost(**values)
+                    inputs, _ = self.get_inputs(rno)
+                    input_iris = [input.output_iri for input in inputs.values()]
+                    owncost = self.cost(self.triplestore, input_iris, self.output_iri)
                     base.cost[i] += owncost
+                    # values = get_values(inputs, rno, magnitudes=True)
```
quaat, line 330: Delete commented-out code

tripper/mappings/mappings.py
```diff
@@ -588,7 +597,10 @@ def getcost(target, stepname):
         cost = soCost.get(target, default_costs[stepname])
         if cost is None:
             return None
-        return function_repo[cost] if cost in function_repo else float(cost)
+        # return function_repo[cost] if cost in function_repo else float(cost)
```
quaat, line 600: Delete commented out code

quaat (changes requested):
Looks ok. Just remove commented out code and merge

tripper/mappings/mappings.py
jesper-friis: done

PR #69, opened 2023-01-25
Add map() method to Triplestore

jesper-friis:
Add a map() method to triplestore that is equivalent to `add_mapsTo()`, except that it reverts the two first arguments and removes the third `property_name` argument.

The motivation for this method is that it has a more natural order of arguments. 

The reason for the reverse argument order of `add_mapsTo()` was to make mappings from data models easier by allowing to provide the data model as one argument and the property name as another. However, with the use of namespaces, it turned out that it was as easy to refer to properties using their full IRI.


tripper/triplestore.py
```diff
@@ -495,6 +495,32 @@ def prefix_iri(self, iri: str, require_prefixed: bool = False):
                 raise NamespaceError(f"No prefix defined for IRI: {iri}")
         return iri
 
+    def map(
```
quaat, line 498: Why is there a method that just calls another method instead of just calling the other method directly?

quaat (approved):

PR #70, opened 2023-01-26
Made the value optional + added some cleanup

jesper-friis:
Support using mappings to build a workflow without executing it. In this case there are no need for explicit values, hence we make them optional.


daniel-sintef (approved):
Please clarify the TODO a bit more before merging.

tripper/mappings/mappings.py
```diff
@@ -336,14 +368,17 @@ def show(
...
-    ):
+    ) -> str:
         """Returns a string representation of the mapping routes to this step.
 
         Arguments:
             routeno: show given route.  The default is to show all routes.
             name: Name of the last mapping step (mainly for internal use).
             indent: How of blanks to prepend each line with (mainly for
-                internal use).
+            internal use).
```
CasperWA, line 378:
```suggestion
                internal use).
```
Indentation is missing here.

Also, the warning message written out by MkDocs should direct one exactly to _this_ location in the code. Let's break it down:

Warning message:

```shell
WARNING  -  mkdocstrings_handlers: tripper.mappings.mappings.MappingStep.show: Failed to get 'name: description' pair from 'internal use).'
```

Now first it's given as a WARNING and it's specifically the `mkdocstrings` extension that's throwing the error. The `mkdocstrings` extension is the one responsible for building the API reference based on the Python docstrings and more.
Then it references a specific method (`show`) in the `MappingStep` class in the `tripper.mappings.mappings` module.
Finally, it seems to not be able to parse a `name: description` pair. This is something that is given normally under a `Parameters:`, `Arguments:`, `Args:`, `Raises:` or similar block. It's essentially saying there's an apparent argument name and description given that it cannot parse from the docstring. The line in question that it cannot parse is also given: `internal use).`. I.e., this is the actual code string piece it tried to parse as a `name: description` pair but failed.

An easy way of debugging this warning is then to search for `internal use).` in the module file and ensure that it's from a doc-string in the `MappingStep.show()` module.
From here, this line is found and it's clear that it should be indented to state that it belongs to the `indent` description and is not a new stand-alone `name: description` pair.

CasperWA (changes requested):

tripper/mappings/mappings.py
jesper-friis: Thanks for solving this. 

It would have been nice with a file:line number, but when you understand the WARNING message it is useful and easy to search for as you say. Another gotcha is that you have to understand that the ERROR is a result of the warning and not something else with a missing error message...

CasperWA: By ERROR here, you mean the error of the workflow failing? If yes, then this is due to this workflow job running `mkdocs build` using the `--strict` option, turning warnings into errors, i.e., returning a non-zero errorcode if a warning happens during the build.
To turn this off (NOT RECOMMENDED) one can set the `warnings_as_errors` input differently in the workflow. See the ci-cd documentation [here](https://sintef.github.io/ci-cd/latest/workflows/ci_tests/#inputs_3).

CasperWA: Also. These messages are produced by MkDocs. I don't think the format can be easily changed. But you're welcome to look into it. The documentation for MkDocs is [here](https://www.mkdocs.org/user-guide/).

CasperWA (approved):

PR #91, opened 2023-03-12
Added add_data(), get_value() and add_interpolation_source() methods to Triplestore

### LINKED ISSUE #90: Transformations based on data sources
Add support for the following use case:

            Assume we have a data source that relates water temperature
            (mapped to EX.Temp) to the amount of blue-green algae (mapped to
            EX.AlgaeConc). By registering it with

            >>> ts.add_interpolation_source(EX.Temp, EX.AlgaeConc)

            we can now ask for the blue-green algae concentration in a fjord,
            given we have a data source with the water temperature field in
            the same fjord.

To implement this we would need little more support to handle data in the tripper API. For this we would need two new methods; an `add_data()` method that register a new data source in the triplestore (as a callback function, such that no data is read at registration time and allows to use partial OTEAPI pipelines) and a `get_value()` method that returns the individual value (for a given mapping route) of an ontological concept.

jesper-friis:
Closes #90 

See #90 for a description of this PR. The aim of this PR is to implement `add_interpolation_source()`, but that required the help from the `add_data()` and `get_value()` methods.

The `add_data()` and `get_value()` methods are designed such that data sources are not accessed before their value are requested. They are also intended to work well with OTEAPI (but for that we should add a convenient layer on top of `add_data()` which will be the subject for a new PR). 

This PR also make it possible to work with actual data via tripper without DLite. If it turns out to work well, DLite should be updated to utilise the new `add_data()` and `get_value()` methods.

## Type of change:
- [ ] Bug fix.
- [x] New feature.
- [ ] Documentation update.


tripper/triplestore.py
```diff
@@ -818,20 +836,260 @@ def _add_function_emmo(self, func, expects, returns, base_iri):
...
-        #    val = f"{func_iri}_output{i+1}"
-        #    self.add((val, RDF.type, DataSet))
-        #    self.add((val, MAP.mapsTo, iri))
-        #    self.add((func_iri, hasOutput, val))
-        # if doc_string:
-        #    self.add((func_iri, DCTERMS.description, en(doc_string)))
-
         return func_iri
+
+    def add_interpolation_source(  # pylint: disable=too-many-arguments
```
quaat, line 841: I don't think we should add this to the production code yet. Let's explore a bit what implications of registering soft sensors will have. We will typically want to add this as part of the domain ontology

quaat (changes requested):
I feel we need a more fundamental discussion here. There are some fundemental differences in using RDF based triplestores and using Knowledge Graphs. The SPARQL/Reasoning framework is excellent for logic, the knowledge graphs are suitable for complex data. For the tasks involving graph traversals etc, query languages such as Gremlin is a lot more suitable (remlin rely on the property graph data model, where data is represented as nodes (or vertices) connected by edges (or relationships). Both nodes and edges can have properties). Gremlin is an imperative graph traversal language designed for property graph databases (e.g., Apache TinkerPop or Neo4j). It supports graph traversal operations and enables users to perform complex queries by chaining together traversal steps.

I feel that here you use the same tool for solving two very different problems. The wrapper layer should clearly distinguish between RDF data models and knowledge graphs, and that is my main point that we need to agree on.

quaat (approved):

tripper/triplestore.py
jesper-friis: > I don't think we should add this to the production code yet. Let's explore a bit what implications of registering soft sensors will have. We will typically want to add this as part of the domain ontology

I agree. That is why I made issue #92 and asked to look at it - the idea with that issue is to keep tripper as a common interface to triplestore backends and move the methods working on the triplestore into a separate (sub)package.

tests/test_get_data.py
```diff
@@ -0,0 +1,73 @@
+"""Test Triplestore.get_data() and related methods."""
```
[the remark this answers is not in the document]
jesper-friis: > I am not sure why we need to mix knowledge graphs and ontologies in the same wrapper utility. We should at least keep them separate in terms of naming and not pretend we suggest to build graphs of data in SPARQL triplestores. There are much better technologies for handling these graphs that use quite different query languages (for instance the gremlin query language).

I agree. That is why I made issue #92 and asked to look at it - the idea with that issue is to keep tripper as a common interface to triplestore backends and move the methods working on the triplestore into a separate (sub)package.

PR #129, opened 2023-08-31
Added support for lists in tripper.convert

jesper-friis:
# Description:
The `save_dict()` and `load_dict()` in the tripper.convert module saves and loads dicts as their name indicate. However, we also need support for lists, since OTEAPI represents mappings as a list in the configurations.

This PR generalises the above functions to also support lists. To reflect that, has the names of the functions also been changed to `save_container()` and `load_container()`. The old function names still exists, but are deprecated.

We have also removed the `bases` argument, since the base (superclass) is now used and can therefore not be allowed to be changed by the user.

A bug in parse_literal() has also been fixed.

## Type of change:

- [x] Bug fix.
- [x] New feature.
- [ ] Documentation update.
- [ ] Testing.

### DIFF
--- a/tripper/convert/convert.py
+++ b/tripper/convert/convert.py
@@ -1,13 +1,24 @@
 """Tripper module for converting between RDF and other repetations."""
 # pylint: disable=invalid-name,redefined-builtin
-from typing import TYPE_CHECKING, Mapping
+import warnings
+from typing import TYPE_CHECKING, Mapping, Sequence
 from uuid import uuid4
 
-from tripper import DCAT, DCTERMS, EMMO, OWL, RDF, RDFS, Literal, Namespace
+from tripper import (
+    DCAT,
+    DCTERMS,
+    EMMO,
+    MAP,
+    OWL,
+    RDF,
+    RDFS,
+    Literal,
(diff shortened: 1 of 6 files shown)


tripper/convert/convert.py
```diff
@@ -35,115 +46,150 @@
...
-                    lang=lang,
-                    recognised_keys=recognised_keys,
-                    keep=keep,
-                )
-            )
-        elif dvalue is None:
-            value = OWL.Nothing
+
+    def get_obj_iri(obj, uuid):
+        """Return IRI for Python object `obj`.  The `uuid` argument is
```
quaat, line 90: Please explain why this method returns either an IRI or a list

quaat (changes requested):

tripper/convert/convert.py
jesper-friis: The nested help-function get_obj_iri() will only return an IRI. I updated its docstring to explain that it also populates the list of triples (built up in the outer function from_container()) with a representation of `obj` .

`from_container()` will always return a list of triples.

quaat (approved):

PR #132, opened 2023-09-09
Cleaned up mappings module

jesper-friis:
# Description:
Minor bugfixes
- fix f-strings in `Value.__repr__()`
- do not bypass mapping-function if number of inputs is one

## Type of change:

- [x] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Testing.

### DIFF
--- a/tripper/mappings/mappings.py
+++ b/tripper/mappings/mappings.py
@@ -270,6 +270,10 @@ class MappingStep:
         Returns:
             Evaluation result.
         """
+        if not self.number_of_routes():
+            raise MissingRelationError(
+                f"no route to evaluate '{self.output_iri}'"
+            )
         if quantity is None:
             quantity = Quantity
         if routeno is None:
--- a/tripper/mappings/mappings.py
+++ b/tripper/mappings/mappings.py
@@ -219,7 +219,7 @@ class MappingStep:
         self.joined_input: "Inputs" = {}
 
     def add_inputs(self, inputs: "Inputs") -> None:
-        """Add input dict for an input route."""
+        """Add input Mapping (e.g., dict) for an input route."""
         assert isinstance(inputs, Mapping)  # nosec B101
         self.input_routes.append(inputs)
 
--- a/tripper/mappings/mappings.py
+++ b/tripper/mappings/mappings.py
@@ -231,7 +231,11 @@ class MappingStep:
         only one input.
 
         If the `join_mode` attribute is true, the input is remembered, but
-        first added when join_input() is called.
+        first added when `join_input()` is called.
+
+        Arguments:
+            input: A mapping step or a value.
+            name: Name assigned to the argument.
         """
         assert isinstance(input, (MappingStep, Value))  # nosec B101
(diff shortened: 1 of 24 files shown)


tripper/mappings/mappings.py
```diff
@@ -277,11 +277,12 @@ def eval(
...
-        if len(inputs) == 1 and all(
-            isinstance(v, Value) for v in inputs.values()
-        ):
-            (value,) = tuple(inputs.values())
-        elif self.function:
+        # if len(inputs) == 1 and all(
+        #    isinstance(v, Value) for v in inputs.values()
+        # ):
+        #    (value,) = tuple(inputs.values())
+        # elif self.function:
```
francescalb, line 284: Why to keep the commented out code?

francescalb (changes requested):
Functionality is being changed without changes in tests. I guess this means that the test was missing previously. Can you please add one?

tripper/mappings/mappings.py
jesper-friis: I would like to make sure that things actually work as expected before removing the commented out lines, since they might address a special case I didn't thought about when creating this PR. The best check for that is DLite PR [#639](https://github.com/SINTEF/dlite/pull/639), which requires that this PR is merged first.

jesper-friis:
> Functionality is being changed without changes in tests. I guess this means that the test was missing previously. Can you please add one?

This PR does not change any functionality, it only fixes some bugs.

sygout:
Mappings are different from dict ? so what is the reason for the change?

jesper-friis:
> Mappings are different from dict ? so what is the reason for the change?

I guess you are referring to line 223. It is just a generalisation, to allow the `input` argument to be any type of mapping, not only dict.

tripper/mappings/mappings.py
```diff
@@ -220,7 +220,7 @@ def __init__(
 
     def add_inputs(self, inputs: "Inputs") -> None:
         """Add input dict for an input route."""
```
ajeklund, line 222:
```suggestion
        """Add input Mapping (e.g., dict) for an input route."""
```

ajeklund (changes requested):
I've added a suggestion and a question, otherwise I think that this PR is fine. I agree that no features have been added.

tripper/mappings/mappings.py
```diff
@@ -220,7 +220,7 @@ def __init__(
...
-        assert isinstance(inputs, dict)  # nosec B101
+        assert isinstance(inputs, Mapping)  # nosec B101
```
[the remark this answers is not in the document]
jesper-friis: We are only using the functionality expected by a mapping (key item access), so it shouldn't cause any problems (unless we have more explicit tests for dict rather than mapping). 

This function is mainly intended for internal use by the mappings module. Otherwise, we should really have documented better how the `inputs` argument should be structured.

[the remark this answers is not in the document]
ajeklund: OK. Thank you!

ajeklund (approved):
Approved. Thank you!

PR #136, opened 2023-09-29
Clearified that tripper does not depend on DLite and Pydantic

jesper-friis:
# Description:
Removed hard requirement on DLite in tests to clarify that tripper does not depend on DLite and Pydantic.

## Type of change:

- [ ] Bug fix.
- [ ] New feature.
- [ ] Documentation update.
- [ ] Testing.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,6 +1,7 @@
 """Test utils"""
 # pylint: disable=invalid-name,too-few-public-methods
-import dlite
+import warnings
+
 import pytest
 
 from tripper import DCTERMS, RDFS, XSD, Literal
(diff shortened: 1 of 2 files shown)


francescalb:
Since we now do not depend on DLite, we should remove Dlite-Python from dependencies. However, as we have a test that checks for something that is included oly when we have Dlite installed I guess we can include DLite-Python in the dev requirements. Furthermore, we should then run the tests both with and without the dev_requirements installed and check that we get the correct output in both cases.

francescalb:
> Since we now do not depend on DLite, we should remove Dlite-Python from dependencies. However, as we have a test that checks for something that is included oly when we have Dlite installed I guess we can include DLite-Python in the dev requirements. Furthermore, we should then run the tests both with and without the dev_requirements installed and check that we get the correct output in both cases.

I need to correct myself: DLite-Python is only in the docs and testing dependencies. I think this is fine then.

PR #163, opened 2024-01-02
Retain datatype when listing literals from rdflib

### LINKED ISSUE #162: Literals are lost when listing triples with rdflib
The following example shows that literals are lost when listing triples with the rdflib backend:

```python
>>> from tripper import Triplestore

>>> ts = Triplestore("rdflib")
>>> ts.parse(format="turtle", data='<http://ex#s> <http://ex#p> "abc"^^<http://www.w3.org/2001/XMLSchema#string> .')

# Calling the graph rdflib object directly, shows that we correctly have parsed the triple 
>>> list(ts.backend.graph.triples((None, None, None)))
[(rdflib.term.URIRef('http://ex#s'),
  rdflib.term.URIRef('http://ex#p'),
  rdflib.term.Literal('abc', datatype=rdflib.term.URIRef('http://www.w3.org/2001/XMLSchema#string')))]

# But the literal is lost when listing with the `triples()` method in Triplestore or the backend:
>>> list(ts.triples())
[('http://ex#s', 'http://ex#p', Literal('abc'))]
>>> list(ts.backend.triples((None, None, None)))
[('http://ex#s', 'http://ex#p', Literal('abc'))]
```

jesper-friis:
# Description
Closes #162 

## Type of change:

- [ ] New feature.
- [ ] Documentation update.
- [x] Testing.

### DIFF
--- a/tripper/mappings/mappings.py
+++ b/tripper/mappings/mappings.py
@@ -122,7 +122,7 @@ class Value:
             args.append(f", property_iri={self.property_iri}")
         if self.cost:
             args.append(f", cost={self.cost}")
-        return f"Value({self._value}{''.join(args)})"
+        return f"Value({self._value!r}{''.join(args)})"
 
     def get_value(self, unit=None, magnitude=False, quantity=None) -> "Any":
         """Returns the evaluated value of given input route number.
--- a/tripper/utils.py
+++ b/tripper/utils.py
@@ -134,11 +134,7 @@ def parse_literal(literal: "Any") -> "Literal":
         lang = literal.language
 
     if not lang and hasattr(literal, "datatype"):
-        datatype = (
-            literal.datatype
-            if literal.__class__ == str
-            else str(literal.datatype)
-        )
+        datatype = str(literal.datatype)
 
     # This will handle rdflib literals correctly and probably most other
     # literal representations as well.
--- a/tripper/utils.py
+++ b/tripper/utils.py
@@ -128,10 +128,18 @@ def parse_literal(literal: "Any") -> "Literal":
     if isinstance(literal, Literal):
         return literal
 
+    if hasattr(literal, "lang"):
+        lang = literal.lang
+    elif hasattr(literal, "language"):
+        lang = literal.language
+
+    if hasattr(literal, "datatype"):
+        datatype = literal.datatype
+
     # This will handle rdflib literals correctly and probably most other
     # literal representations as well.
     if hasattr(literal, "value"):
-        return Literal(literal.value)
+        return Literal(literal.value, lang=lang, datatype=datatype)
 
     if not isinstance(literal, str):
         if isinstance(literal, tuple(Literal.datatypes)):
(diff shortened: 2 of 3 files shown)


tripper/mappings/mappings.py
```diff
@@ -122,7 +122,7 @@ def __repr__(self):
             args.append(f", property_iri={self.property_iri}")
         if self.cost:
             args.append(f", cost={self.cost}")
-        return f"Value({self._value}{''.join(args)})"
+        return f"Value({self._value!r}{''.join(args)})"
```
ajeklund, line 125: Is this change related to the handling of literals?

tripper/utils.py
```diff
@@ -128,10 +128,22 @@ def parse_literal(literal: "Any") -> "Literal":
...
 
+    if hasattr(literal, "lang"):
+        lang = literal.lang
+    elif hasattr(literal, "language"):
+        lang = literal.language
+
+    if not lang and hasattr(literal, "datatype"):
+        datatype = (
+            literal.datatype
+            if literal.__class__ == str
```
ajeklund, line 139:
```suggestion
            if literal.datatype.__class__ == str
```
(I'm not sure about this, please check.)

tripper/utils.py
```diff
@@ -128,10 +128,22 @@ def parse_literal(literal: "Any") -> "Literal":
...
+            else str(literal.datatype)
+        )
+
     # This will handle rdflib literals correctly and probably most other
     # literal representations as well.
     if hasattr(literal, "value"):
-        return Literal(literal.value)
+        return Literal(literal.value, lang=lang, datatype=datatype)
```
ajeklund, line 146: The rdflib documentation states that `Literal` contains a literal value and either a datatype or language tag (i.e. not both). [Link](https://rdflib.readthedocs.io/en/stable/rdf_terms.html) (see the first paragraph)

Is it still correct to supply both?

ajeklund (commented):
Looks good! I added a couple of questions.

tripper/mappings/mappings.py
jesper-friis: Not directly. The change was done to improve the representation of Value objects, needed while developing/debugging this PR.

tripper/utils.py
jesper-friis: `tripper.Literal` and `rdflib.Literal` are two different things. Like rdflib, tripper does not expect language and datatype to be provided at the same time, so typically either `lang` or `datatype` is None.

tripper/utils.py
ajeklund: No need for that change.

What I don't understand is the logic: why do you want to assign `literal.datatype` straight off only when `literal` is a `str`, and `str(literal.datatype)` for any other case?

jesper-friis: You are completely right. There is no need for that conditional, we can just `str(literal.datatype)` for all cases. The reason we need the `str()` function is that `literal` may be an instance of `rdflib.Literal` whos `datatype` attribute is not a string, but a `rdflib.URIRef` object.

ajeklund: Thank you!

ajeklund (approved):
Thank you for fixing this!

Issue #161, opened 2024-01-02
Surprising literal comparisons


jesper-friis:
The following comparisons between a Literal and another object may be surprising:

```python
In [7]: Literal("abc", datatype=XSD.string) == "abc"
Out[7]: True

In [10]: Literal(1, datatype=XSD.int) == 1
Out[10]: False
 
In [11]: Literal(1, datatype=XSD.int) == "1"
Out[11]: True
```

This is an unintended consequence of `Literal` being a subclass of `str`. 

A suggestion is to add an `__eq__()` method to `Literal` that uses `tripper.utils.parse_literal()` to convert all non-literal objects to a Literal and then require that value, lang and datatype all must be equal for equality to be true.

jesper-friis:
Adding `__eq__()` (and `__hash__()`) to `Literal` causes failures in mapping due to the changed comparison logics.

Updating mappings to the new behaviour of literal comparisons may require little more effort. Are there also some third party libraries that depends on tripper that we will break?

PR #165, opened 2024-01-03
Retain literal types in collection backend

### LINKED ISSUE #160: Retain literal types in collection backend
The PR https://github.com/SINTEF/dlite/pull/755 adds datatype (and language) for literals to DLite collections. This allows to store a knowledge base in a collection and get it back without loss of type information.

Update the collection backend to utilise the above PR in DLite to allow storing and retrieving triples without loss of type information.

jesper-friis:
# Description:
Closes #160 

Note that this PR builds on top of PR #161 

## Type of change:

- [x] New feature.
- [ ] Documentation update.
- [ ] Testing.

### DIFF
--- a/tripper/backends/collection.py
+++ b/tripper/backends/collection.py
@@ -57,19 +57,49 @@ class CollectionStrategy:
 
     def triples(self, triple: "Triple") -> "Generator[Triple, None, None]":
         """Returns a generator over matching triples."""
-        for s, p, o in self.collection.get_relations(*triple):
-            yield s, p, parse_object(o)
+        for s, p, o, d in self.collection.get_relations(*triple, rettype="T"):
+            if d:
+                lang = d[1:] if d[0] == "@" else None
+                dt = None if lang else d
+                yield s, p, Literal(o, lang=lang, datatype=dt)
+            else:
+                yield s, p, o
+
+            # lang = d[1:] if d and d[0] == "@" else None
+            # dt = None if lang else d
+            # print("*** d:", d)
+            # yield s, p, Literal(o, lang=lang, datatype=dt)
 
     def add_triples(self, triples: "Sequence[Triple]"):
         """Add a sequence of triples."""
         for s, p, o in triples:
             v = parse_object(o)
-            v_str = v.n3() if isinstance(v, Literal) else v
-            self.collection.add_relation(s, p, v_str)
+            o = v if isinstance(v, str) else v.value
+            d = (
+                None
+                if not isinstance(v, Literal)
+                else f"@{v.lang}"
+                if v.lang
+                else v.datatype
+            )
+            # d = f"@{v.lang}" if v.lang else v.datatype
+            # v_str = v.n3() if isinstance(v, Literal) else v
+            print(f"*** add_triple: v={repr(v)}, o={o}, d={d}")
+            self.collection.add_relation(s, p, o, d)
+        print("-----")
+        # for spod in self.collection.get_relations=
 
     def remove(self, triple: "Triple"):
         """Remove all matching triples from the backend."""
         s, p, o = triple
         v = parse_object(o)
-        v_str = v.n3() if isinstance(v, Literal) else v
-        self.collection.remove_relations(s, p, v_str)
+        o = v if isinstance(v, str) else v.value
+        d = (
+            None
+            if isinstance(v, str)
+            else f"@{v.lang}"
+            if v.lang
+            else v.datatype
+        )
+        # v_str = v.n3() if isinstance(v, Literal) else v
+        self.collection.remove_relations(s, p, o, d)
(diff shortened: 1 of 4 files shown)


tripper/triplestore.py
```diff
@@ -294,33 +294,35 @@ def parse(
...
             source: File-like object or file name.
             format: Needed if format can not be inferred from source.
-            backend: If given, use the parse() method of the specified
-                backend instead of the current backend.
-            backend_kwargs: Dict with additional keyword arguments passed
-                when instantiating `backend`.
+            fallback_backend: If the current backend doesn't implement
+                parse, use the `fallback_backend` instead.
+            fallback_backend_kwargs: Dict with additional keyword arguments
+                for initialising `fallback_backend`.
```
francescalb, line 309: I do not see where these are tested. According to codecov, they are, so currently it should be OK. But it is important that this is tested on purpose and not "by accident", e.g. by using a backend that right now does not contain parse but might in the future.

tripper/triplestore.py
```diff
@@ -342,28 +344,28 @@ def serialize(
...
             format: Format to serialise as.  Supported formats, depends on
                 the backend.
-            backend: If given, use the parse() method of the specified
-                backend instead of the current backend.
-            backend_kwargs: Dict with additional keyword arguments passed
-                when instantiating `backend`.
+            fallback_backend: If the current backend doesn't implement
+                serialisation, use the `fallback_backend` instead.
+            fallback_backend_kwargs: Dict with additional keyword arguments
+                for initialising `fallback_backend`.
```
francescalb, line 350: Same comment as for parse above

tripper/backends/collection.py
```diff
@@ -66,31 +66,37 @@ def triples(self, triple: "Triple") -> "Generator[Triple, None, None]":
...
             # has been declared to return the union of `str` and
             # `Literal`.
             v = parse_object(o)
-            o = v if isinstance(v, str) else v.value
+
+            # Possible bug in mypy
+            # parse_object() is declared to return "Union[str, Literal]".
+            # Dispite of that complains mypy about that `v` has no
+            # attribute "value"
+            obj = v if isinstance(v, str) else str(v.value)  # type: ignore
```
francescalb, line 84:
```suggestion
            obj = v if isinstance(v, str) else str(v.value)
```

I have thested locally and with a draft pull request, I do not get any errors with mypy. Do you have an updated version?

tripper/backends/collection.py
```diff
@@ -66,31 +66,37 @@ def triples(self, triple: "Triple") -> "Generator[Triple, None, None]":
...
             )
-            self.collection.add_relation(s, p, o, d)
+            self.collection.add_relation(s, p, obj, d)
 
     def remove(self, triple: "Triple"):
         """Remove all matching triples from the backend."""
         s, p, o = triple
         v = parse_object(o)
-        o = v if isinstance(v, str) else v.value
+        obj = v if isinstance(v, str) else str(v.value)  # type: ignore
```
francescalb, line 96:
```suggestion
        obj = v if isinstance(v, str) else str(v.value)
```
see comment above

tripper/triplestore.py
```diff
@@ -294,33 +294,35 @@ def parse(
         self,
         source=None,
         format=None,
-        backend=None,
-        backend_kwargs=None,
+        fallback_backend="rdflib",
+        fallback_backend_kwargs=None,
```
francescalb, line 298: Since we didn't need to change backend/backend_kwargs to fallback_backend/fallback_backend_kwargs in the tests, it seems that we do not have adeguate tests.

tripper/backends/collection.py
jesper-friis: The problem came when I updated mypy. I got annoyed with it and changed the return type of utils.parse_object() from `Union[str, Literal]` to `Union[str, Any]` to get rid of these errors...

The source of the problem is that mypy suddenly has got the idea that `Literal` has got a special meaning, since it was added to `typing`. 

It even doesn't understand if I change the return type of utils.parse_object() to `Union[str, tripper.Literal]`. 

Using `Any` in the return works around the problem, so I have removed the comment and the `# type: ignore` comments.

tripper/triplestore.py
jesper-friis: Well, the collection backend does not provide a parse() method, so it will use the fallback backend. That is tested in tests/backends/test_collection.py.

tripper/triplestore.py
jesper-friis: They are tested in tests/backends/test_collection.py

tripper/triplestore.py
jesper-friis: Tested in tests/backends/test_collection.py

tripper/triplestore.py
francescalb: OK, I prefer tests to be a bit more explicit.

francescalb (approved):
Seems OK, although I do not understand the mypy issue.

PR #187, opened 2024-03-03
Correctly convert rdflib bnodes back to tripper

jesper-friis:
# Description:
Correctly convert rdflib bnodes back to tripper

## Type of change:
- [ ] New feature.
- [ ] Documentation update.
- [x] Testing.

### DIFF
--- a/tests/backends/test_rdflib.py
+++ b/tests/backends/test_rdflib.py
@@ -5,13 +5,15 @@ Most of the rdflib backend is already tested in tests/test_triplestore.py.
 
 import pytest
 
-from tripper import Literal, Triplestore
+from tripper import RDFS, Literal, Triplestore
 
 rdflib = pytest.importorskip("rdflib")
 
 
-# Test for issue #162: Literals are lost when listing triples with rdflib
 ts = Triplestore("rdflib")
+EX = ts.bind("", "http://ex#")
+
+# Test for issue #162: Literals are lost when listing triples with rdflib
 ts.parse(
     format="turtle",
     data=(
@@ -26,3 +28,8 @@ assert list(ts.triples()) == [
         Literal("abc", datatype="http://www.w3.org/2001/XMLSchema#string"),
     )
 ]
+
+
+# Test for bnode
+ts.add_triples([("_:bn1", RDFS.subClassOf, EX.s)])
+assert ts.value(predicate=RDFS.subClassOf, object=EX.s) == "_:bn1"
--- a/tests/backends/test_rdflib.py
+++ b/tests/backends/test_rdflib.py
@@ -13,7 +13,7 @@ rdflib = pytest.importorskip("rdflib")
 ts = Triplestore("rdflib")
 EX = ts.bind("", "http://ex#")
 
-# Test for issue #162: Literals are lost when listing triples with rdflib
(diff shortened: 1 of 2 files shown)


tests/backends/test_rdflib.py
```diff
@@ -5,13 +5,15 @@
...
+from tripper import RDFS, Literal, Triplestore
 
 rdflib = pytest.importorskip("rdflib")
 
 
-# Test for issue #162: Literals are lost when listing triples with rdflib
 ts = Triplestore("rdflib")
+EX = ts.bind("", "http://ex#")
+
+# Test for issue #162: Literals are lost when listing triples with rdflib
```
francescalb, line 16: This comment does not make sense. Are we testing that triples listed with rdflib includes literals?

jesper-friis: Agree, the comment is badly phrased. Rephrased it to:

    # Test for issue #162: Literals are lost when accessing triples with rdflib

tests/backends/test_rdflib.py
```diff
@@ -5,13 +5,15 @@
...
+from tripper import RDFS, Literal, Triplestore
 
 rdflib = pytest.importorskip("rdflib")
 
 
-# Test for issue #162: Literals are lost when listing triples with rdflib
 ts = Triplestore("rdflib")
+EX = ts.bind("", "http://ex#")
+
+# Test for issue #162: Literals are lost when accessing triples with rdflib
```
francescalb, line 16: I still think this explanation is strange. Issue 162 is closed, so I do not understand. 
Maybe then just write "check that Literals are kept when accessing triples with rdflib"?
Also, put it just above the assert statement.

tests/backends/test_rdflib.py
```diff
@@ -26,3 +28,8 @@
         Literal("abc", datatype="http://www.w3.org/2001/XMLSchema#string"),
     )
 ]
+
+
+# Test for bnode
+ts.add_triples([("_:bn1", RDFS.subClassOf, EX.s)])
+assert ts.value(predicate=RDFS.subClassOf, object=EX.s) == "_:bn1"
```
francescalb, line 35: We should have blank node and normal node tests for both subject and objects.

francescalb (changes requested):
Please look at the comments

tests/backends/test_rdflib.py
jesper-friis: This was a test added while debugging issue 162. When we have it, I think it is nice to keep the reference to the issue (it is still accessible even though it has been closed). It makes it possible to check the issue if someone would like to understand the reason for adding this test. 

The ts.parse() and assert statements belongs together. The assert statement checks what was added with the parse, so for me the placement of the comment feels logical.

tests/backends/test_rdflib.py
jesper-friis: Agree, we should ideally have more tests here. But normal nodes are already well testet in all the other tests. 

This bnode test, is really the key test for this PR. I added it to this file, because the issue we are trying to solve is specific to how bnodes are retrieved from the rdflib backend.

francescalb (approved):
OK, we will have to agree to disagree.

PR #195, opened 2024-03-10
Enhanced the use of namespaces

### LINKED ISSUE #194: Add caching and extension to tripper.Namespace
`tripper.Namespace` already allow working with ontologies using numerical IDs, like EMMO, in a user friendly way. However, it requires that the ontology is downloaded at every invocation, which may have a detrimental impact on performance. 

Also, some ontologies does not load all its modules by default. This is e.g. the case of EMMO that for instance doesn't load pre-fixed and non-coherent units by default. Hence, by default not all concepts are accessible from the namespace object.

To address these two issues:
1. cache the downloaded ontology
2. make it easy to load additional modules into the namespace

jesper-friis:
# Description
Simplified and improved the usability of namespaces
* Added persistent cache to Namespace. Loading a large namespace like EMMO from cache is now almost instant.
* Added extend_namespace() utility function.
* Simplified the implementation of Namespace, including:
  - replaced `cachemode` argument with the simple boolean argument `reload` 
  - removed `triplestore_url` argument
  - added `format` argument for supporting loading ontologies in formats other than turtle
* Updated the `Triplestore.bind()` method to make it simpler to use.
* Updated the tutorial and improved the description of how namespaces can be used.
* Added and improved tests. Added example for simplified SPARQL queries using tripper Namespace.

Closes #194 

## Type of change
- [x] New feature
- [x] Documentation update
- [x] Testing

## Checklist for the reviewer


docs/tutorial.md
```diff
@@ -33,29 +41,46 @@ Namespace also supports access by label and IRI checking.
...
-It is needed, because the 'rdflib' backend is currently not able to load EMMO from the `http://emmo.info/emmo#` namespace.
 
 ```python
 >>> EMMO = ts.bind(
-...     "emmo", "http://emmo.info/emmo#",
+...     "emmo", "https://w3id.org/emmo#",
 ...     label_annotations=True,
 ...     check=True,
-...     triplestore_url="https://emmo-repo.github.io/versions/1.0.0-beta4/emmo-inferred.ttl",
 ... )
```
francescalb, line 52: This is a nice feature that is good to document in this simple example.
Don't remove it completely

docs/tutorial.md
```diff
@@ -95,14 +120,62 @@ For example:
...
+
 
-Finally Triplestore has two specialised methods `add_mapsTo()` and `add_function()` that simplify working with mappings.
-`add_mapsTo()` is convinient for defining new mappings:
+Writing SPARQL queries using Tripper
+------------------------------------
+A challenge with ontologies using numerical IRIs is that SPARQL queries become difficult to read and understand.
+This challenge is greatly mitigated by using the `label_annotations` feature of Tripper namespaces.
+The example below shows how to write and execute a SPARQL query with Tripper that finds the IRI and unit symbol of all length units.
+Note how:
```
francescalb, line 131:
```suggestion
Note:
```

docs/tutorial.md
```diff
@@ -95,14 +120,62 @@ For example:
...
+1. EMMO classes and properties are written as `{EMMO.LengthUnit}`, which would expand to `https://w3id.org/emmo#EMMO_b3600e73_3e05_479d_9714_c041c3acf5cc`.
+2. That the curly brackets after the `WHERE` clause has to be written `{{`, `}}` because the query is an f-string.
```
francescalb, line 133:
```suggestion
2. The curly brackets after the `WHERE` clause have to be written `{{`, `}}` because the query is an f-string.
```

tests/test_namespace.py
```diff
@@ -63,3 +55,22 @@ def test_namespaces(get_ontology_path: "Callable[[str], Path]") -> None:
         FOOD.NonExisting  # pylint: disable=pointless-statement
 
     assert FOOD2.NonExisting == FOOD2 + "NonExisting"
+
+
+# if True:
+def test_namespace_emmmo():
```
francescalb, line 61:
```suggestion
def test_namespace_emmo():
```

tripper/namespace.py
```diff
@@ -197,6 +258,10 @@ def __eq__(self, other):
 TIME = Namespace("http://www.w3.org/2006/time#")
 FNO = Namespace("https://w3id.org/function/ontology#")
 
+# EMMO = Namespace("http://w3id.org/emmo#")
+# MAP = Namespace("http://w3id.org/domain-mappings#")
+# DM = Namespace("http://w3id.org/datamodel#")
+# OTEIO = Namespace("http://w3id.org/oteio#")
```
francescalb, line 264: When should these be updated? Why not already?

francescalb (changes requested):
Nice, I especially like to better introduction. However, there are quite a few lines not tested. I think these should be added.

docs/tutorial.md
jesper-friis: It is not removed. I only removed the `triplestore_url` argument, since it is not needed any more. All the labels are now fetch from the IRI "https://w3id.org/emmo#" provided as the second argument. 

Enabling simplifications like this, is part of the goal of this PR :-)

tripper/namespace.py
jesper-friis: It is definitely something we should do, but I think it is better to do it in a separate PR, since it may have some implications on existing tests in other repositories.

It is probably a good idea to make a release before and after changing these.

We must also rename the datamodel and oteio repos to domain-datamodel and domain-oteio for the w3id redirections to work.

docs/tutorial.md
francescalb: It is removed in the tutorial. I think it should stay, either in the same example or add another just after to point out this very important possiblity.

jesper-friis: Added a reformulation of the comment about the use of the `triplestore` argument below the code example.

jesper-friis:
> Nice, I especially like to better introduction. However, there are quite a few lines not tested. I think these should be added.

Added some more tests, but:
- No tests are added for the deprecated arguments in Namespace.__init__() (they would have to be removed soon again).
- Windows- and Mac-specific stuff in get_cachedir() are not testet either.

tripper/backends/rdflib.py
```diff
@@ -66,14 +66,12 @@ class RdflibStrategy:
...
         self,
-        base_iri: "Optional[str]" = None,
+        base_iri: "Optional[str]" = None,  # pylint: disable=unused-argument
         database: "Optional[str]" = None,
         triplestore_url: "Optional[str]" = None,
         format: "Optional[str]" = None,  # pylint: disable=redefined-builtin
         graph: "Graph" = None,
     ) -> None:
-        if base_iri:
-            warnings.warn("base_iri", UnusedArgumentWarning, stacklevel=3)
```
francescalb, line 76: Why remove this warning?

francescalb (approved):
Nice, but please explain in the Docstring why base_iri is a valid Argument

PR #196, opened 2024-03-10
Use Python 3.9 in all CI/CD workflows

### LINKED ISSUE #193: Documentation CD (still) fails due to old python version
This is a follow-up issue on issue #191 with PR #192, which _should_ have solved the problem.

The problem is that the "External / Deploy 'latest' documentation" job fails because it tries to use 'mkdocs-material' version higher than 9.5.5 on python 3.7, which is incompatible.

Does this happen because the `publish-package-and-docs` job uses `SINTEF/ci-cd/.github/workflows/cd_release.yml@v1`? Seems like the latest version of that is v2.7.4.

CasperWA:
# Description
Use Python 3.9 throughout all CI/CD workflows.

Fixes #193 

## Type of change
- [ ] New feature
- [ ] Documentation update
- [x] Testing

## Checklist for the reviewer

### DIFF
--- a/.github/workflows/ci_cd_updated_main.yml
+++ b/.github/workflows/ci_cd_updated_main.yml
@@ -18,7 +18,7 @@ jobs:
       update_docs: true
       update_python_api_ref: true
       update_docs_landing_page: true
-      python_version: "3.7"
+      python_version: "3.9"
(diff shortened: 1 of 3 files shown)


ajeklund (approved):
Thank you!

PR #223, opened 2024-06-17
Added an additional test for convert

jesper-friis:
# Description
Added a test to try to catch the issue seen in SS3_wrappers

## Type of change
- [ ] New feature
- [ ] Documentation update
- [x] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/convert/test_convert.py
+++ b/tests/convert/test_convert.py
@@ -5,6 +5,7 @@
 import pytest
 
 
+# if True:
 def test_convertions():
     """Test convertions. Uses rdflib as triplestore backend"""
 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci_tests.yml
```diff
@@ -34,11 +34,14 @@ jobs:
       #  https://pyup.io/vulnerabilities/CVE-2021-41495/44715/
       #  https://pyup.io/vulnerabilities/CVE-2021-41496/44716/
       #  https://pyup.io/vulnerabilities/CVE-2021-34141/44717/
+      # 70612: Jinja2 vulnerability. Only used as subdepndency for mkdocs++ in tripper.
```
francescalb, line 37:
```suggestion
      # 70612: Jinja2 vulnerability. Only used as subdependency for mkdocs++ in tripper.
```

francescalb (changes requested):
Codecov fails now.

More importantly, I think we should have better documentation on how to use these functionalities. Since we cannot do everything in one go, I suggest adding exmamples for how to use (and how it is usful to use it) in load_container. If we make an effort to update documentation every time we make an upgrade and fix we can reach our goal of good documentation.

Also, please descrive what  this PR is about. This might be useful in the documentation update.

francescalb (approved):
OK, this works nicely.

PR #225, opened 2024-06-21
Get restrictions as dicts

jesper-friis:
# Description
Added a `return_dicts` keyword argument to the `Triplestore.retrictions()` method, to allow to get the restrictions as dicts instead of blank nodes. 

Other changes:
* Also renamed the `target` argument of `Triplestore.add_retriction()` and `Triplestore.retrictions()` to `value` in order to be consistent with the naming in EMMOntoPy. These two methods are a very new addition to tripper that haven't been used yet. This change that this change will break existing code is therefore very small.
* Added tests.
* Added documentation in the tutorial.

**Question**
Should we change the default of `return_dicts` to be true? If so, this is the right time to do it, before people starts to use these methods.

## Type of change
- [x] New feature
- [x] Documentation update
- [x] Testing

## Checklist for the reviewer

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -240,8 +240,8 @@ With the `return_dicts` argument set to false, you will instead get an iterator 
 
 ```python
 >>> g = ts.restrictions(cls=ONTO.Bacteria, property=EMMO.hasPart, return_dicts=False)
->>> list(g)  # doctest: +ELLIPSIS
-['_:...']
+>>> next(g) == iri
+True
 
 ```
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -229,7 +229,7 @@ It returns an iterator over all restrictions that matches the provided criteria.
 For example:
 
 ```python
->>> g = ts.restrictions(cls=ONTO.Bacteria, property=EMMO.hasPart, return_dicts=True)
+>>> g = ts.restrictions(cls=ONTO.Bacteria, property=EMMO.hasPart, asdict=True)
 >>> list(g)  # doctest: +ELLIPSIS
 [{'iri': '_:...', 'cls': 'http://example.com/onto#Bacteria', 'property': 'https://w3id.org/emmo#EMMO_17e27c22_37e1_468c_9dd7_95e137f73e7f', 'type': 'exactly', 'cardinality': 1, 'value': 'http://example.com/onto#Cell'}]
 
@@ -238,7 +238,7 @@ For example:
 With the `return_dicts` argument set to false, an iterator over the IRIs of all matching restrictions is returned:
 
 ```python
->>> g = ts.restrictions(cls=ONTO.Bacteria, property=EMMO.hasPart, return_dicts=False)
+>>> g = ts.restrictions(cls=ONTO.Bacteria, property=EMMO.hasPart, asdict=False)
 >>> next(g) == iri
 True
 
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -187,13 +187,12 @@ True
 ```
 
 ### Class restrictions
-When working with OWL ontologies, you often need to look up or add class restrictions.
-A [restriction] restricts a class to only those individuals that satisfy the restriction.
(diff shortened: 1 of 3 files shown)


docs/tutorial.md
```diff
@@ -186,6 +186,66 @@ True
 
 ```
 
+### Class restrictions
+When working with OWL ontologies, you often need to look up or add class restrictions.
+A [restriction] restricts a class to only those individuals that satisfy the restriction.
+The Triplestore class has two convenient methods for this, that don't require knowledge about how restrictions are represented in RDF.
+They only support basic restrictions without any nested logical constructs.
+For more advanced restrictions, we recommend to use [EMMOntoPy] or [Owlready2].
```
francescalb, line 194:
```suggestion
When working with OWL ontologies, it is often required to inspect or add class [restriction]s.
The Triplestore class has two convenient methods for this, that do not require knowledge about how restrictions are represented in RDF.
Only support basic restrictions, without any nested logical constructs, are supoprted.
For more advanced restrictions, we recommend to use [EMMOntoPy] or [Owlready2].
```

docs/tutorial.md
```diff
@@ -186,6 +186,66 @@ True
...
+
+A restriction is described by the following set of parameters.
```
francescalb, line 196:
```suggestion
A [restriction] is described by the following set of parameters.
```

docs/tutorial.md
```diff
@@ -186,6 +186,66 @@ True
...
+    - *min*: minimum cardinality restriction (target is a class IRI)
+    - *max*: maximum cardinality restriction (target is a class IRI)
+    - *value*: Value restriction (target is an IRI of an individual or a literal)
+
+  * **cardinality**: the cardinality value for cardinality restrictions.
+  * **value**: The IRI or literal value of the restriction target.
+
+For example. Lets assume that you have a class `onto:Bacteria` that you want to logically restrict to be unicellular.
+In Manchester syntax, this can be stated as `onto:Bacteria emmo:hasPart exactly 1 onto:Cell`.
+With tripper, you can state it with
```
francescalb, line 213:
```suggestion
As an example, the class `onto:Bacteria` can be logically restricted to be unicellular.
In Manchester syntax, this can be stated as `onto:Bacteria emmo:hasPart exactly 1 onto:Cell`.
With Tripper this can be stated as:
```

docs/tutorial.md
```diff
@@ -186,6 +186,66 @@ True
...
+...     type="exactly",
+...     cardinality=1,
+...     value=ONTO.Cell,
+... )
+
+```
+The returned `iri` is the blank node IRI of the new restriction.
+
+
+To find the above restriction, you can use the `restrictions()` method.
```
francescalb, line 228:
```suggestion
To find the above restriction, the `restrictions()` method can be used.
```

docs/tutorial.md
```diff
@@ -186,6 +186,66 @@ True
...
+For example:
+
+```python
+>>> g = ts.restrictions(cls=ONTO.Bacteria, property=EMMO.hasPart, return_dicts=True)
+>>> list(g)  # doctest: +ELLIPSIS
+[{'iri': '_:...', 'cls': 'http://example.com/onto#Bacteria', 'property': 'https://w3id.org/emmo#EMMO_17e27c22_37e1_468c_9dd7_95e137f73e7f', 'type': 'exactly', 'cardinality': 1, 'value': 'http://example.com/onto#Cell'}]
+
+```
+
+With the `return_dicts` argument set to false, you will instead get an iterator over the IRIs of all matching restrictions:
```
francescalb, line 239:
```suggestion
With the `return_dicts` argument set to false, an iterator over the IRIs of all matching restrictions is returned:
```

francescalb (changes requested):
Some suggested changes to the tutorial, but only wording. It looks very good!

I think we should change to return_dicts = True as default as also changing target to value is a breaking change. We just have to state very clearly in the release that this is done, and that thisis a breaking change on a a feature that was added in the previous realease.

francescalb (approved):
Very nice.

PR #229, opened 2024-06-29
Added a fix for parsing rdflib literals.

jesper-friis:
# Description
We cannot use the `value` attribute of an `rdflib.Literal` to access its value, since it is `None` for some datatypes (at least for rdf:JSON). 
Seems more robust to use the existence of the `n3()` method to test for a literal and the string representation to access its value.

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tripper/utils.py
+++ b/tripper/utils.py
@@ -237,7 +237,7 @@ def parse_literal(literal: "Any") -> "Any":
     # This should handle rdflib literals correctly (and probably most other
     # literal representations as well)
     if hasattr(literal, "value"):
-        # Note that in rdflib 6.3, the `value` attribute may None for some
+        # Note that in rdflib 6.3, the `value` attribute may be None for some
         # datatypes (like rdf:JSON) even though a non-empty value exists.
         # As a workaround, we use the string representation if the value
         # attribute is None.
--- a/tripper/utils.py
+++ b/tripper/utils.py
@@ -234,10 +234,9 @@ def parse_literal(literal: "Any") -> "Any":
     ):
         datatype = str(literal.datatype)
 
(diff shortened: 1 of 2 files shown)


tripper/utils.py
```diff
@@ -234,10 +234,16 @@ def parse_literal(literal: "Any") -> "Any":
     ):
         datatype = str(literal.datatype)
 
-    # This will handle rdflib literals correctly and probably most other
-    # literal representations as well.
+    # This should handle rdflib literals correctly (and probably most other
+    # literal representations as well)
     if hasattr(literal, "value"):
-        return Literal(literal.value, lang=lang, datatype=datatype)
+        # Note that in rdflib 6.3, the `value` attribute may None for some
```
francescalb, line 240:
```suggestion
        # Note that in rdflib 6.3, the `value` attribute may be None for some
```

tripper/utils.py
```diff
@@ -234,10 +234,16 @@ def parse_literal(literal: "Any") -> "Any":
     ):
         datatype = str(literal.datatype)
 
-    # This will handle rdflib literals correctly and probably most other
-    # literal representations as well.
+    # This should handle rdflib literals correctly (and probably most other
+    # literal representations as well)
     if hasattr(literal, "value"):
```
francescalb, line 239: But can a literal not have the attribute? Isn't it always there, but sometimes with value None? (I thought the reason for this fix is exactly that)

jesper-friis: Agree, the argument name `literal` is not very well chosen. It was changed by Casper from `value` to `literal` in one of his early reviews.

francescalb (approved):

PR #247, opened 2024-09-02
Added argument `ignore_unrecognised` to tripper.convert()

jesper-friis:
# Description
Added argument `ignore_unrecognised` to tripper.convert().

**Question**: One could consider to set `ignore_unrecognised=True` by default. It will be a small change in behavior, but more user friendly. I think that would be wise change.

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/.github/workflows/cd_release.yml
+++ b/.github/workflows/cd_release.yml
@@ -6,9 +6,9 @@ on:
     - published
 
 jobs:
-  publish-package-and-docs:
+  build:
     name: External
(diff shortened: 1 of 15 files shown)


francescalb:
I have trouble seeing how to use this function prperly. Could you add some examples?
The reason that I am asking for this as a part of this PR is that it is hard to really see what is happening. Especially since a test has not been added or updated.

jesper-friis:
The example is now added to module docstring. It is checked by doctest.

francescalb (approved):
very nice.

PR #252, opened 2024-09-24
Added option to Triplestore.value() to return a generator over all matching values

jesper-friis:
# Description
Added option to Triplestore.value() to return a generator over all matching values. Also added more tests.

**Question**: Is there a better way to indicate that we should return all matches? Using `any=None` to indicate that is not very intuitive. But using a special value of the `any` argument is logical, since returning any match (`any=True`), returning all matches (`any=None`) and require only one match (`any=False`) are mutually exclusive.

## Type of change
- [x] New feature
- [ ] Documentation update
- [x] Testing

## Checklist for the reviewer

### DIFF
--- a/tripper/triplestore.py
+++ b/tripper/triplestore.py
@@ -581,10 +581,10 @@ class Triplestore:
             object: Possible criteria to match.
             default: Value to return if no matches are found.
             any: Used to define how many values to return. Can be set to:
-                'False' (default): return the value or raise UniquenessError
+                `False` (default): return the value or raise UniquenessError
                 if there is more than one matching value.
-                 'True': return any matching value if there is more than one.
-                 'None': return a generator over all matching values.
+                `True`: return any matching value if there is more than one.
+                `None`: return a generator over all matching values.
             lang: If provided, require that the value must be a localised
                 literal with the given language code.
 
@@ -625,7 +625,7 @@ class Triplestore:
         except StopIteration:
             return value
 
-        if any:
+        if any is True:
             return value
         raise UniquenessError("More than one match")
 
--- a/tripper/triplestore.py
+++ b/tripper/triplestore.py
@@ -580,8 +580,10 @@ class Triplestore:
             predicate: Possible criteria to match.
             object: Possible criteria to match.
             default: Value to return if no matches are found.
-            any: If true, return any matching value, otherwise raise
-                UniquenessError.
+            any: If True, return any matching value if there is more than one.
+                 If False, raise UniquenessError if there is more than one
+                 matching value.
+                 If None, return a generator over all matching values.
             lang: If provided, require that the value must be a localised
                 literal with the given language code.
 
@@ -607,25 +609,34 @@ class Triplestore:
             for triple in triples:
                 value = triple[idx]
                 if isinstance(value, Literal) and value.lang == lang:
-                    if any:
+                    if any is True:
                         return value
+                    if any is None:
+                        yield value
                     if first:
                         raise UniquenessError("More than one match")
                     first = value
             if first is None:
                 return default
+        elif any is None:
+            for triple in triples:
+                print("xxx0")
+                yield triple[idx]
         else:
             try:
                 triple = next(triples)
             except StopIteration:
+                print("xxx1")
                 return default
 
         try:
             next(triples)
         except StopIteration:
+            print("xxx2")
             return triple[idx]
 
         if any:
+            print("xxx3")
             return triple[idx]
         raise UniquenessError("More than one match")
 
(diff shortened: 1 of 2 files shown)


tripper/triplestore.py
```diff
@@ -600,33 +602,30 @@ def value(  # pylint: disable=redefined-builtin
...
+        try:
+            value = next(triples)[idx]
+        except StopIteration:
+            return default
 
         try:
             next(triples)
         except StopIteration:
-            return triple[idx]
+            return value
```
francescalb, line 625: What is value here?

tripper/triplestore.py
```diff
@@ -600,33 +602,30 @@ def value(  # pylint: disable=redefined-builtin
...
 
         if any:
-            return triple[idx]
+            return value
```
francescalb, line 628: What is value here?

francescalb:
> **Question**: Is there a better way to indicate that we should return all matches? Using `any=None` to indicate that is not very intuitive. But using a special value of the `any` argument is logical, since returning any match (`any=True`), returning all matches (`any=None`) and require only one match (`any=False`) are mutually exclusive.


How about using matches="all", "any", "unique" and have unique as default? I suppose it needs to be added in addition to any, to be deprecated.

tripper/triplestore.py
jesper-friis: It is assigned on line 618

tripper/triplestore.py
jesper-friis: > What is value here?

value is assigned on line 618

jesper-friis:
> > **Question**: Is there a better way to indicate that we should return all matches? Using `any=None` to indicate that is not very intuitive. But using a special value of the `any` argument is logical, since returning any match (`any=True`), returning all matches (`any=None`) and require only one match (`any=False`) are mutually exclusive.
> 
> How about using matches="all", "any", "unique" and have unique as default? I suppose it needs to be added in addition to any, to be deprecated.

I like that much better, but the method is modelled after the `rdflib.Graph.value()` method that has `any` as a boolean argument. We could deprecate `any` and add `matches` as a new argument. But having an interface that follows rdflib makes it very easy to replace direct use of rdflib with tripper in existing applications. Thoughts?

tripper/triplestore.py
```diff
@@ -580,8 +580,10 @@ def value(  # pylint: disable=redefined-builtin
             predicate: Possible criteria to match.
             object: Possible criteria to match.
             default: Value to return if no matches are found.
-            any: If true, return any matching value, otherwise raise
-                UniquenessError.
+            any: If True, return any matching value if there is more than one.
+                 If False, raise UniquenessError if there is more than one
+                 matching value.
+                 If None, return a generator over all matching values.
```
francescalb, line 586:
```suggestion
            any: Used to define how many values to return. Argument:
                'False' (default): return the value or raise UniquenessError 
                if there is more than one matching value.             
                 'True': return any matching value if there is more than one.
                 'None': return a generator over all matching values.
```

francescalb:
> > > **Question**: Is there a better way to indicate that we should return all matches? Using `any=None` to indicate that is not very intuitive. But using a special value of the `any` argument is logical, since returning any match (`any=True`), returning all matches (`any=None`) and require only one match (`any=False`) are mutually exclusive.
> > 
> > 
> > How about using matches="all", "any", "unique" and have unique as default? I suppose it needs to be added in addition to any, to be deprecated.
> 
> I like that much better, but the method is modelled after the `rdflib.Graph.value()` method that has `any` as a boolean argument. We could deprecate `any` and add `matches` as a new argument. But having an interface that follows rdflib makes it very easy to replace direct use of rdflib with tripper in existing applications. Thoughts?

I ma not sure, I guess it depends how important it is to stay close to rdflib. I tried to make the doicstring more understandable, but I am not sure if it worked.

jesper-friis:
> > > > **Question**: Is there a better way to indicate that we should return all matches? Using `any=None` to indicate that is not very intuitive. But using a special value of the `any` argument is logical, since returning any match (`any=True`), returning all matches (`any=None`) and require only one match (`any=False`) are mutually exclusive.
> > > 
> > > 
> > > How about using matches="all", "any", "unique" and have unique as default? I suppose it needs to be added in addition to any, to be deprecated.
> > 
> > 
> > I like that much better, but the method is modelled after the `rdflib.Graph.value()` method that has `any` as a boolean argument. We could deprecate `any` and add `matches` as a new argument. But having an interface that follows rdflib makes it very easy to replace direct use of rdflib with tripper in existing applications. Thoughts?
> 
> I ma not sure, I guess it depends how important it is to stay close to rdflib. I tried to make the doicstring more understandable, but I am not sure if it worked.

Thanks for the improved docstring. I think that it is a good idea to stay close to rdflib, to not increase the barrier for people to use tripper in existing applications. It is still possible to add a `matches` argument in a future PR that, if given, overrides `any` without breaking backward compatibility.

francescalb (approved):

PR #255, opened 2024-10-13
Allowing backends to indicate whether they prefer the sparql interface

jesper-friis:
# Description
Allow backends to add a class attribute `prefer_sparql` attribute, indicating  whether they prefer calls to the SPARQL `query()` method instead of the `triples()` method. 

Added the `Triplestore.prefer_sparql` property for easy access to this attribute. 

Also renamed TriplestoreError to TripperError and some additional cleanup of warnings.

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer


tripper/triplestore.py
```diff
@@ -539,6 +539,14 @@ def list_databases(cls, backend: str, **kwargs):
     # interfaces to the triples(), add_triples() and remove() methods
     # implemented by all backends.
 
+    prefer_sparql = property(
+        fget=lambda self: getattr(self.backend, "prefer_sparql", None),
+        doc=(
+            "Whether the backend prefer SPARQL over the triples() interface. "
+            "Is None if not specified by the backend."
```
francescalb, line 546:
```suggestion
            "Whether the backend uses SPARQL and not the triples() interface. "
            "Is None if not specified by the backend."
```

Actually I do not entirely understand what this is. Can you exlpain it better so we can improve the docstring?

jesper-friis: This is really about preferences. There is a big difference between backends, some can be used very efficient with Triplestore.triples() (e.g. collection, ontopy, ...) while other are very efficient with Triplestore.query() (e.g. fuseki, graphdb, ...). 

Even though Tripper requires that the Triplestore.triples() is implemented, we have to use Triplestore.query() for some backends (like fuseki) because of the issue with blank nodes getting renamed. The purpose of this PR is to allow a user of tripper to automatically select the most appropriate interface depending on the current backend.

francescalb: This you can write in the docstring.

jesper-friis: done

tripper/triplestore.py
```diff
@@ -539,6 +539,23 @@ def list_databases(cls, backend: str, **kwargs):
...
+            "\n\n"
+            "Even though Tripper requires that the Triplestore.triples() is "
+            "implemented, we have to use Triplestore.query() for some "
+            "backends in some cases (like fuseki when working on RDF lists) "
+            "because of the issue with blank nodes getting renamed. "
+            "\n\n"
+            "The purpose of this property is to allow a user of tripper to "
+            "automatically select the most appropriate interface depending "
+            "on the current backend."
```
francescalb, line 555:
```suggestion
            "Even though Tripper requires that the Triplestore.triples() is "
            "implemented, Triplestore.query() must be used for some "
            "backends in specific cases (like fuseki when working on RDF lists) "
            "because of how blank nodes are treated. "
            "\n\n"
            "The purpose of this property is to allow a user of tripper to "
            "automatically select the most appropriate interface depending "
            "on the current backend."
```

francescalb (commented):
Do we have any documentation anywhere on what is needed to create a new backend? If so we should add there that the prefer_sparql attribute should be set. If not, we should have an issue on making such a template/documentation.

Furthermore, have you tested this with PyBackTrip? I guess this is not backwards compatible?

jesper-friis:
> Do we have any documentation anywhere on what is needed to create a new backend? If so we should add there that the prefer_sparql attribute should be set. If not, we should have an issue on making such a template/documentation.
> 
> Furthermore, have you tested this with PyBackTrip? I guess this is not backwards compatible?

The best documentation is the interface.py file. I have already added the `prefer_sparql` atteibute to that file.

francescalb:
> > Do we have any documentation anywhere on what is needed to create a new backend? If so we should add there that the prefer_sparql attribute should be set. If not, we should have an issue on making such a template/documentation.
> > Furthermore, have you tested this with PyBackTrip? I guess this is not backwards compatible?
> 
> The best documentation is the interface.py file. I have already added the `prefer_sparql` atteibute to that file. The interface.py file is not used for anything else apart from documentation.


Great, then I think it would be very good to add a sentence about looking into this file if one wants to make a new backend. (In the documentation for developers part). It is enough with one sentence for now

jesper-friis:
> > > Do we have any documentation anywhere on what is needed to create a new backend? If so we should add there that the prefer_sparql attribute should be set. If not, we should have an issue on making such a template/documentation.
> > > Furthermore, have you tested this with PyBackTrip? I guess this is not backwards compatible?
> > 
> > 
> > The best documentation is the interface.py file. I have already added the `prefer_sparql` atteibute to that file. The interface.py file is not used for anything else apart from documentation.
> 
> Great, then I think it would be very good to add a sentence about looking into this file if one wants to make a new backend. (In the documentation for developers part). It is enough with one sentence for now

done

PR #256, opened 2024-10-13
 Dataset

jesper-friis:
# Description
Added new `tripper.dataset` module for documenting datasets.

All the main features documented in the module docstring have now been implemented.

The tests currently fails because `emmo:hasDimension` has been removed from EMMO. How should we deal with this?

For the next PRs, we should:
- [ ] Add some kind of session with a user-specific information that should not be stored in the triplestore, including:
      - credentials like username and password to external storages
      - local root for relative paths
      - default storage (e.g. redis) to work against during execution of the user workflow 
  Exactly how this session should work is still unclear. It will probably be a class with the functions documented in the
  top of the dataset module as methods + some additional methods for dealing with the internal state.
- [ ] Decide whether we want to move the protocol plugins from DLite to Tripper? It may also be possible to create a new package.
- [ ] Update CONTEXT_URL to point to master branch (or to github pages if we copy the context there)
- [ ] Create a good example/tutorial
- [ ] Document the `@context`
- [ ] Fix the `__TODO__`'s in the code

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer


tests/input/datasets.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+    # List of consumers of this dataset
+    datasink:
+        # This simple consumer wants the data in png format
+      - storeURL: mydata.png
+        mediaType: image/png
+        generator:
+          functionType: application/vnd.dlite-generate
+          configuration:
+            driver: image
```
francescalb, line 47: Why should a dataset have a list of consumers? Isn't one of the core ideas of this documentation framework that datasources and datasinks should be documented independently?

tests/input/semdata.yaml
```diff
@@ -0,0 +1,80 @@
...
+    configuration:
+      driver: hitachi
+
+
+generators:
+  - "@id": gen:sem_hitachi
+    "@type": oteio:Generator
+    generatorType: application/vnd.dlite-generate
+    configuration:
+      driver: hitachi
```
francescalb, line 80: What is the generator for an already defined /existing dataset?

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
+"""Module for documenting datasets with Tripper.
+
+The dataset documentation follows the DCAT structure and is exposed as
+Python dicts with attribute access in this module.  This dict
```
francescalb, line 4: I suggest adding a link to DCAT here, and also explicitly stating that the statements included in the context.json are supported (if I have understood it correctly).

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+    else functools.lru_cache(maxsize=1)
+)
+
+
+# Pytest can't cope with this
+# EMMO = Namespace(
+#     iri="https://w3id.org/emmo#",
+#     label_annotations=True,
+#     check=True,
+# )
```
francescalb, line 60: What do we need this for? EMMO is defined as namespace here https://github.com/EMMC-ASBL/tripper/blob/1d38c2c8daafd9d4d63a099e9171cf49c9b2083f/tripper/namespace.py#L287

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+    dataset: "Optional[Union[str, dict]]" = None,
+    distribution: "Optional[Union[str, dict]]" = None,
+    generator: "Optional[str]" = None,
+    prefixes: "Optional[dict]" = None,
+    use_sparql: "Optional[bool]" = None,
+) -> None:
+    """Saves a documented dataset to a data resource.
+
+    Arguments:
+        ts: Triplestore to load data from.
```
francescalb, line 114: Triple store to load data documentation or data? Hvere does the data come from?

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+        data: Bytes representation of the dataset to save.
+        class_iri: IRI of a class in the ontology (e.g. an `emmo:DataSet`
+            subclass) that describes the dataset that is saved.
+            Is used to select the `distribution` if that is not given.
+            If `distribution` is also given, a
+            `dcat:distribution value <distribution>` restriction will be
+            added to `class_iri`
+        dataset: IRI of dataset for the data to be saved.
+            Or a dict with additional documentation of the dataset.
```
francescalb, line 123: This is confusing. What is the IRI. Does it already exist? Or is this something we have to invent/set here.

Also why are we allowed to set an IRI or give additional information, should it not be possible to give both simultaneously?

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+        distribution: IRI of distribution for the data to be saved.
+            Or a dict additional documentation of the distribution,
+            like media type, parsers, generators etc...
```
francescalb, line 126: Same comment as above.

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+) -> None:
+    """Saves a documented dataset to a data resource.
+
+    Arguments:
+        ts: Triplestore to load data from.
+        data: Bytes representation of the dataset to save.
+        class_iri: IRI of a class in the ontology (e.g. an `emmo:DataSet`
+            subclass) that describes the dataset that is saved.
+            Is used to select the `distribution` if that is not given.
+            If `distribution` is also given, a
```
francescalb, line 119: https://www.w3.org/TR/vocab-dcat-3/#Property:dataset_distribution

Is it this distribution?

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+    if gen and "configuration" in gen and "options" in gen.configuration:
+        # __TODO__: allow options to also be a dict
+        options.append(gen.configuration["options"])
+    id = (
+        distribution["accessService"].get("identifier")
+        if "accessService" in distribution
+        else None
+    )
+    with Protocol(scheme, location, options=";".join(options)) as pr:
+        pr.save(data, id)
```
francescalb, line 239: Seems strange that ther are not other, more persistent packages we can use for this.

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+def load(
+    ts: Triplestore,
+    iri: str,
+    distributions: "Optional[Union[str, Sequence[str]]]" = None,
+    use_sparql: "Optional[bool]" = None,
+) -> bytes:
+    """Load dataset with given IRI from its source.
+
+    Arguments:
+        ts: Triplestore to load data from.
```
francescalb, line 258: Load data or documentation of data`?

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+
+    Returns:
+        An updated copy of `dct`.
+
+    Notes:
+        The keys in `dct` and `kwargs` may be either properties defined in the
+        [JSON-LD context](https://raw.githubusercontent.com/EMMC-ASBL/oteapi-dlite/refs/heads/rdf-serialisation/oteapi_dlite/context/0.2/context.json)
+        or one of the following special keywords:
+          - "@id": Dataset IRI.  Must always be given.
+          - "@type": IRI of the ontology class for this type of data.
```
francescalb, line 346:
```suggestion
          - "@id": Dataset IRI.  Must always be given. (Note as "_id" in key)
          - "@type": IRI of the ontology class for this type of data. (Note: as "_type" in key)
```

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+    statements = get_values(dct, "statements")
+    statements.extend(get_values(dct, "mappings"))
+    ts.add_triples(statements)
+
+    # Save data models
+    datamodels = get_values(dct, "datamodel")
+    try:
+        # pylint: disable=import-outside-toplevel
+        import dlite
+        from dlite.dataset import add_dataset
```
francescalb, line 395: Why are datamodels dependent on dlite? DataModels should be independent of the python (or c) api.

tripper/dataset/dataset.py
jesper-friis: The problem is that currently

```python
>>> from tripper import EMMO
>>> EMMO.Atom
'https://w3id.org/emmo#Atom'
```

which is not the correct IRI for the Atom class. It includes the numerical UUID. 

The above way to define the EMMO namespace produces the correct result, but doesn't work with pytest.

tripper/dataset/dataset.py
jesper-friis: True, the data is not stored in the triplestore. The text should be updated

tests/input/semdata.yaml
francescalb: I think we should not have both parsers and generators for the same thing.

Parsers only for existing data or for data that we know are outputs of specific models.

Generators only for data that does not exist, e.g. inputs to models. 

Parsers and generators in the same file that is supposed to document an existing dataset is very confusing.

jesper-friis: Good point. I removed the generator from the distribution of the SEM dataset. 

Do you also think it is confusing to keep the declaration of the generator in the YAML file? 
If this file is about documentation of existing datasets, the generator should not be documented in this file. On the other hand, it is easier to manage if corresponding parsers and generators are are documented next to each other.

One thing I did not add in the YAML file was the `sem:SEMImage dcat:distribution value semdata:newdistr` value restriction, where `semdata:newdistr` is the distribution that will be generated.

tripper/dataset/dataset.py
jesper-friis: done

tripper/dataset/dataset.py
jesper-friis: done

tests/input/semdata.yaml
francescalb: I think Generator and Parsers are typically two very different things. Yes, in our simple examples there is a one to one correpondence. But in most real cases I think that a parser will only capture a subset of all the information available in the dataset (i.e. all the things that are present in the data model, but not all). The generator can only generate a dataset based on what the datamodel describes. 

In this specific case, the generator will never generate a SEM-image, and therefore I think it is confusing to keep in here. 

We have to be very clear about the difference between documenting datasets that have been created by exp or models, and generating new datasets based on (DLite) datamodels. In the first case it is really new, independent data. In the second case it is data that is fetched via a datamodel and repurposed (transformed) to a format (distribution) that we wish for.

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+
+@cache  # type: ignore
+def get_context(timeout: float = 5, fromfile: bool = True) -> dict:
+    """Return context as a dict.
+
+    Arguments:
+        timeout: Number of seconds before timing out.
+        fromfile: Whether to load the context from local file.
+
+    Whether to read the context from file."""
```
francescalb, line 549:
```suggestion
    """
```

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+    elif isinstance(data, list):
+        for ele in data:
+            if isinstance(ele, (dict, list)):
+                values.extend(get_values(ele, key))
+    return values
+
+
+@cache  # type: ignore
+def get_context(timeout: float = 5, fromfile: bool = True) -> dict:
+    """Return context as a dict.
```
francescalb, line 543: Which context is it getting? All context? Context defined how? This needs to be described better.

tests/dataset/test_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def test_get_context():
+    """Test get_context()."""
+    from tripper.dataset import get_context
+
+    context = get_context()
+    assert isinstance(context, dict)
+    assert "@version" in context
+    assert len(context) > 20
```
francescalb, line 25: I do not understand where the context is found. It is not referring to any triplestore as far as I can see?

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+            (isinstance(v, str) and not v.endswith(("#", "/")))
+            or isinstance(v, dict)
+        )
+    }
+    shortnames.setdefault(RDF.type, "@type")
+    return shortnames
+
+
+def load_list(ts: Triplestore, iri: str):
+    """Load and return RDF list whos first node is `iri`."""
```
francescalb, line 589: I do not understand the purpose of this function.

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+                add_dataset(ts, dm)
+            else:
+                try:
+                    dm = dlite.get_instance(uri)
+                except (
+                    dlite.DLiteMissingInstanceError  # pylint: disable=no-member
+                ):
+                    # __FIXME__: check session whether want to warn or re-reise
+                    # in this case
+                    warnings.warn(f"cannot load datamodel: {uri}")
```
francescalb, line 419:
```suggestion
```

Remove. This function should save the datadocumentation to the triplestore. It should not check whether the datamodel is available. And definitely not wheter it is a valid dlite datamodel.

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,942 @@
...
+        from dlite.dataset import add_dataset
+    except ModuleNotFoundError:
+        if datamodels:
+            warnings.warn(
+                "dlite is not installed - data models will not be added to "
+                "the triplestore"
+            )
+    else:
+        for url in get_values(dct, "datamodelStorage"):
+            dlite.storage_path.append(url)
```
francescalb, line 404:
```suggestion
            dlite.storage_path.append(url)
```

If you realyl ant to check something you can check that the url to the datamodel exists. But we should not in any way be dependent on dlite here.

tripper/dataset/dataset.py
francescalb: OK, I just find it strange that the datamodel should go into the same triplestore (necessarily).

tests/input/semdata.yaml
```diff
@@ -0,0 +1,80 @@
...
+datasets:
+  - "@id": semdata:SEM_cement_batch2/77600-23-001/77600-23-001_5kV_400x_m001
+    "@type": sem:SEMImage
+    title: SEM image of cement
+    description: Back-scattered SEM image of cement sample 77600 from Heidelberg, polished with 1 µm diamond compound.
+    creator: Sigurd Wenner
+    contactPoint: Sigurd Wenner <Sigurd.Wenner@sintef.no>
+    inSeries: semdata:SEM_cement_batch2/77600-23-001
+
+    datamodel: http://onto-ns.com/meta/characterisation/0.1/SEMImage
```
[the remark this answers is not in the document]
jesper-friis: Updated semdata.yml based on feedback from @francescalb.

tests/input/semdata.yaml
```diff
@@ -0,0 +1,80 @@
...
+    statements:
+      - ["@id", "sem:fromSample", "sample:SEM_cement_batch2/77600-23-001"]
+      - ["@id", "emmo:isDescriptionOf", "mat:concrete1"]
+
+    distribution:
+      #downloadURL: sftp://nas.aimen.es/P_MATCHMAKER_SHARE_SINTEF/SEM_cement_batch2/77600-23-001/77600-23-001_5kV_400x_m001.tif
+      downloadURL: https://github.com/EMMC-ASBL/tripper/raw/refs/heads/dataset/tests/input/77600-23-001_5kV_400x_m001.tif
+      mediaType: image/tiff
+      parser: parser:sem_hitachi
+      generator: gen:sem_hitachi
```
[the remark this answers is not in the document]
francescalb: OK, after a discussion we have agreed to continue with datamodel at the dataset-level. If a different datamodel is used to describe the same raw data it will be a different dataset. However, I still think we should _NOT_ have a generator when describig this particular dataset as it is not a datasink. We should have a separate description of a possible datasink. A desitribution description of a datasink can have both parser and a generator. The generator is necessary. The parser optional, and is intended to be used to document the generated dataset as a dataset after it has been generated.

tripper/triplestore.py
```diff
@@ -159,6 +160,26 @@ def __init__(
...
+            base_iri: Assigned to the `base_iri` argument.
+            closed: Whether the triplestore is closed.
+            kwargs: Dict with additional keyword arguments.
+            namespaces: Dict mapping namespace prefixes to IRIs.
+            package: Name of Python package if the backend is implemented as
+                a relative module. Assigned to the `package` argument.
+
+        Notes:
+            If the backend establish a connection that should be closed,
+            it is useful to instantiate the Triplestore as a context manager:
```
francescalb, line 174:
```suggestion
            If the backend establishes a connection that should be closed
            it is useful to instantiate the Triplestore as a context manager:
```

tripper/triplestore.py
```diff
@@ -159,6 +160,26 @@ def __init__(
...
+
+                >>> import tripper
+                >>> with tripper.Triplestore("rdflib") as ts:
+                ...     print(ts.backend_name)
+                rdflib
+
+            This will ensure that the connection is closed when the scope of
+            the with-statement is left.
```
francescalb, line 182:
```suggestion
            This ensures that the connection is automatically closed when the context manager exits.
```

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,964 @@
+# pylint: disable=line-too-long
+"""Module for documenting datasets with Tripper.
+
+The dataset documentation follows the [DCAT] structure and is exposed
+as Python dicts with attribute access in this module.  The semantic
+meaning of the keywords in this dict are defined by a [JSON-LD context].
+
+High-level functions for accessing and storing actual data:
+  - `load()`: Load documented dataset from its source.
+  - `save()`: Save documented dataset to a data resource.
```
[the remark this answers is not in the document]
jesper-friis: I do not disagree. For now, lets move load() and save() to a separate module.

jesper-friis:
> We have not discussed and agreed that the methods intended for dataset loading and fethcing from a database should be in a different module. Please separate out those methods and write in the moduel documentation that these are likely to be moved out of tripper at a later stage, since tripper is dedicated to the interaction with a triplestore, i.e. det datadocumentation and not the raw data itself.

OK. Will do...

tripper/dataset/dataset.py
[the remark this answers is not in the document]
jesper-friis: I have now separated out dataaccess.py from dataset.py

jesper-friis:
> > We have not discussed and agreed that the methods intended for dataset loading and fethcing from a database should be in a different module. Please separate out those methods and write in the moduel documentation that these are likely to be moved out of tripper at a later stage, since tripper is dedicated to the interaction with a triplestore, i.e. det datadocumentation and not the raw data itself.
> 
> OK. Will do...

done

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,736 @@
...
+    return dct
+
+
+def _load_sparql(ts: Triplestore, iri: str) -> dict:
+    """Load `iri` from triplestore by calling `ts.query()`."""
+    # The match-all pattern `(:|!:)*` in the query string below
+    # ensures that the returned triples includes nested structures,
+    # like distributions in a dataset. However, it does not include
+    # references to named resources, like parsers and generators.
+    # This is good, because it limits the number of returned triples.
```
francescalb, line 264:
```suggestion
    # This choice was made because it limits the number of returned triples.
```

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,736 @@
...
+    # The `recur()` function will load such named resources recursively.
+    #
+    # Note that this implementation completely avoid querying for
```
francescalb, line 267:
```suggestion
    # Note that this implementation completely avoids querying for
```

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,736 @@
...
+    return prepare_datadoc(d)
+
+
+def save_datadoc(
+    ts: Triplestore, file_or_dict: "Union[str, Path, dict]"
+) -> dict:
+    """Populate triplestore with data documentation.
+
+    Arguments:
+        ts: Triplestore to load dataset from.
```
francescalb, line 453:
```suggestion
        ts: Triplestore to load dataset documentation from.
```

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,736 @@
...
+                ts.add_triples(ts2.triples())
+
+    # Add statements and datamodels to triplestore
+    save_extra_content(ts, d)
+    return d
+
+
+def prepare_datadoc(datadoc: dict) -> dict:
+    """Return an updated version of dict `datadoc` that is prepared for
+    serialisation to RDF."""
```
francescalb, line 491: Prepare with what? Please specify so the documentation is complete for a new user.

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,736 @@
...
+    return pipeline
+
+
+def list_dataset_iris(ts: Triplestore, **kwargs):
+    """Return a list of IRIs for all datasets matching a set of criterias
+    specified by `kwargs`.
+
+    Arguments:
+        ts: Triplestore to search.
+        kwargs: Match criterias.
```
francescalb, line 710: It is difficult for a new user to understand how this can be used. How about providing and example?

tripper/triplestore.py
```diff
@@ -160,6 +161,26 @@ def __init__(
             kwargs: Keyword arguments passed to the backend's __init__()
                 method.
 
+        Attributes:
+            backend_name: Name of backend.
+            base_iri: Assigned to the `base_iri` argument.
+            closed: Whether the triplestore is closed.
```
francescalb, line 167: What does this mean? can we have an example?

tripper/triplestore.py
```diff
@@ -160,6 +161,26 @@ def __init__(
...
+            kwargs: Dict with additional keyword arguments.
```
francescalb, line 168: kwargs should be at the end.

tripper/dataset/dataset.py
```diff
@@ -0,0 +1,736 @@
...
+            (isinstance(v, str) and not v.endswith(("#", "/")))
+            or isinstance(v, dict)
+        )
+    }
+    shortnames.setdefault(RDF.type, "@type")
+    return shortnames
+
+
+def load_list(ts: Triplestore, iri: str):
+    """Load and return RDF list whos first node is `iri`."""
```
francescalb, line 383:
```suggestion
    """Load and return RDF list whose first node is `iri`."""
```

francescalb (approved):
Approved provided that _all_ issues discussed are added as well-described issues in the issue tracker, and that the requested docstring updates are done to make the code useful for external users.

tripper/dataset/dataset.py
jesper-friis: Updated the documentation. The function is mostly ment for internal use. Only the function listed at the top of the module are actually intended for users.

tripper/dataset/dataset.py
jesper-friis: Good point. Done

tripper/triplestore.py
jesper-friis: These are class attributes. Documented using the google docstring style: https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html#example-google

It means that you can do the following

```python
>>> ts = Triplestore("rdflib")
>>> ts.closed
False
>>> ts.close()
>>> ts.closed
True

```

tripper/triplestore.py
jesper-friis: `kwargs` is an attribute assigned to the `kwargs` argument. Hence `ts.kwargs` is a dict with additional keyword arguments.

francescalb (approved):
Approved, but please make the issues in the issuetracker

tripper/context/0.2/context.json
```diff
@@ -0,0 +1,144 @@
...
+    "status": "adms:status",
+    "versionNotes": "adms:versionNotes",
+    "distribution": {
+      "@id": "dcat:distribution",
+      "@type": "@id"
+    },
+    "contactPoint": "dcat:contactPoint",
+    "hasCurrentVersion": "dcat:hasCurrentVersion",
+    "hasVersion": "dcat:hasVersion",
+    "inSeries": "dcat:inSeries",
```
torhaugl, line 34:
```suggestion
    "inSeries": {
      "@id" : "dcat:inSeries",
      "@type" : "@id"
    },
```

PR #261, opened 2024-11-11
Predefined EMMO namespace with checking and label lookup

jesper-friis:
# Description
Enable checking and label lookup by default for the EMMO namespace. The following now works out of the box

```python
>>> from tripper import EMMO
>>> EMMO.Atom
'https://w3id.org/emmo#EMMO_eb77076b_a104_42ac_a065_798b2d2809ad'

```

Also implemented lazy loading of the triplestore when a `Namespace` with checking or label lookup is created. This ensures that we still can import tripper without having rdflib installed, although `EMMO.Atom` will fail with a `ModuleNotFoundError`. 

Added workaround for a design bug in pytest - without the completely unmotivated special handling of the `__wrapped__` attribute pytest would always fail during its collection phase. We should consider to replace pytest with a more decent test framework...

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/test_namespace.py
+++ b/tests/test_namespace.py
@@ -93,8 +93,9 @@ def test_triplestore_arg() -> None:
     with pytest.raises(NoSuchIRIError):
         FAM.NonExisting  # pylint: disable=pointless-statement
 
-    with pytest.raises(NamespaceError):
-        Namespace("http://example.com", check=True, triplestore=Ellipsis)
+    EX = Namespace("http://example.com", check=True, triplestore=Ellipsis)
+    with pytest.raises(NamespaceError):  # lazy loading of namespace
+        EX.Class  # pylint: disable=pointless-statement
 
 
 # if True:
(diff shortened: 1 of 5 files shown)


torhaugl (approved):
This looks OK to me. I tested the function on my PC and it works (but rdflib must be installed as stated in the PR). The EMMO namespace was moved from namespace.py to __init__.py, was this intentional? If so, go ahead and merge

jesper-friis:
> This looks OK to me. I tested the function on my PC and it works (but rdflib must be installed as stated in the PR). The EMMO namespace was moved from namespace.py to **init**.py, was this intentional? If so, go ahead and merge

Yes, this was intentional. Creating the EMMO namespace in the `namespace` module would lead to cyclic dependencies during import time, since creating a namespace with checking requires `triplestore`, which imports `namespace`. We avoid this cyclic dependency by creating EMMO in `__init__` after both namespace and triplestore are imported. 

Merging

PR #266, opened 2024-11-23
Add support for Python 3.13

jesper-friis:
# Description
Add support for Python 3.13.

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -138,7 +138,7 @@ good-names = [
 [tool.pytest.ini_options]
 minversion = "7.0"
 #addopts = "-rs --cov=tripper --cov-report=term --doctest-modules"
-addopts = "-rs --doctest-modules --ignore=examples"
+addopts = "-rs --doctest-modules --ignore=examples --ignore=tripper"
 filterwarnings = [
     "ignore:.*imp module.*:DeprecationWarning",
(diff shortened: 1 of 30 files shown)


pyproject.toml
```diff
@@ -136,7 +138,7 @@ good-names = [
 
 [tool.pytest.ini_options]
 minversion = "7.0"
-addopts = "-rs --cov=tripper --cov-report=term --doctest-modules"
+addopts = "-rs --cov=tripper --cov-report=term --doctest-modules --ignore=examples --ignore=tripper"
```
francescalb, line 141:
```suggestion
addopts = "-rs --cov=tripper --cov-report=term --doctest-modules --ignore=examples"
```
I also do not like ignoring of examples, but tripper which is our code-base should definitely not be ignored.

jesper-friis: Replaced it with new `--doctest-ignore-import-errors`

jesper-friis: > Replaced it with new `--doctest-ignore-import-errors`

jesper-friis: > I also do not like ignoring of examples, but tripper which is our code-base should definitely not be ignored.

Replaced it with new `--doctest-ignore-import-errors`

PR #272, opened 2024-12-15
Updated dataset module

jesper-friis:
# Description
Updated dataset module including the following changes:
  - Allow to add other types of entries to the triplestore that are not datasets. Ex: samples, models, instruments, people, projects...
  - Renamed list_data_iris() to search_iris(). It can now be use to search for all types of entries. **Question**: Would `search()` be an even better name?
  - Renamed prepare() to as_jsonld() and made it part of the public API

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/input/semdata.yaml
+++ b/tests/input/semdata.yaml
@@ -1,4 +1,5 @@
 ---
+
 # This extends the list of prefixes that are already defined in the context
 prefixes:
   sem: https://w3id.com/emmo/domain/sem/0.1#
@@ -62,10 +63,6 @@ datasets:
       downloadURL: sftp://nas.aimen.es/P_MATCHMAKER_SHARE_SINTEF/SEM_cement_batch2
       mediaType: inode/directory
 
-  - "@id": sample:SEM_cement_batch2/77600-23-001
-    "@type": chameo:Sample
-    title: Series for SEM images for sample 77600-23-001.
-
 
 parsers:
   - "@id": parser:sem_hitachi
@@ -81,3 +78,10 @@ generators:
     generatorType: application/vnd.dlite-generate
     configuration:
       driver: hitachi
+
+
+# Other entities, like samples, instruments, persons, models etc...
+other_entries:
+  - "@id": sample:SEM_cement_batch2/77600-23-001
+    "@type": chameo:Sample
+    title: Series for SEM images for sample 77600-23-001.
--- a/tripper/dataset/dataset.py
+++ b/tripper/dataset/dataset.py
@@ -9,12 +9,13 @@ High-level function for populating the triplestore from YAML documentation:
   - `save_datadoc()`: Save documentation from YAML file to the triplestore.
 
 Functions for searching the triplestore:
-  - `list_dataset_iris()`: Get IRIs of matching datasets.
+  - `search_iris()`: Get IRIs of matching entries in the triplestore.
 
 Functions for working with the dict-representation:
   - `read_datadoc()`: Read documentation from YAML file and return it as dict.
   - `save_dict()`: Save dict documentation to the triplestore.
   - `load_dict()`: Load dict documentation from the triplestore.
+  - `as_jsonld()`: Return the dict as JSON-LD (represented as a Python dict)
 
 Functions for interaction with OTEAPI:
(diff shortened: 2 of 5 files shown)


tripper/dataset/dataset.py
```diff
@@ -199,8 +206,7 @@ def save_extra_content(ts: Triplestore, dct: dict) -> None:
                 except (
                     dlite.DLiteMissingInstanceError  # pylint: disable=no-member
                 ):
-                    # __FIXME__: check session whether want to warn or re-reise
-                    # in this case
+                    # __FIXME__: check session whether to warn or re-reise
```
francescalb, line 209:
```suggestion
                    # __FIXME__: check session whether to warn or re-raise
```

tests/input/semdata.yaml
```diff
@@ -81,3 +78,10 @@ generators:
     generatorType: application/vnd.dlite-generate
     configuration:
       driver: hitachi
+
+
+# Other entities, like samples, instruments, persons, models etc...
+other_entries:
+  - "@id": sample:SEM_cement_batch2/77600-23-001
+    "@type": chameo:Sample
+    title: Series for SEM images for sample 77600-23-001.
```
francescalb, line 87: I find it a bit strange that crucial information like @id and @type end up under other_entries. What is the rationale for this?

tripper/dataset/dataset.py
```diff
@@ -120,14 +126,15 @@ def save_dict(
...
     Notes:
         The keys in `dct` and `kwargs` may be either properties defined in the
-        [JSON-LD context](https://raw.githubusercontent.com/EMMC-ASBL/oteapi-dlite/refs/heads/rdf-serialisation/oteapi_dlite/context/0.2/context.json)
-        or one of the following special keywords:
+        [JSON-LD context] or one of the following special keywords:
           - "@id": Dataset IRI.  Must always be given.
           - "@type": IRI of the ontology class for this type of data.
             For datasets, it is typically used to refer to a specific subclass
             of `emmo:DataSet` that provides a semantic description of this
             dataset.
```
francescalb, line 134: This does not render well in the documentation. Check out howt o make the lists inthe notes.

tests/input/semdata.yaml
jesper-friis: Yes, other_entries is not a good label. My simpleminded initial idea was that we are only documenting datasets. But that is not true. We also want to represent samples, models, people, instruments, projects, etc in the knowledge base. These are not datasets but other types of entries. 

A better label would be "samples". We could easily add support for that, but we would probably need a more general framework where the user can extend the categories. Ideally such extensions should be done with user-supplied JSON-LD context. A good solution should probably go into a separate PR.

tripper/dataset/dataset.py
jesper-friis: Thanks. Done

torhaugl (approved):
For me this PR looks reasonable. It is mostly renaming `prepare` to `as_jsonld`, and adding a possibility to add `other_entries` such as samples. I am still of the opinion that context should be put into the yaml file, and that we should use JSON-LD or (YAML-LD)[https://json-ld.github.io/yaml-ld/spec/] directly, which would immeadiately add the `other_entries` functionality in a more "intuitive" way. That discussion is better in a different PR I think. If you @francescalb dont have any more comments, feel free to merge

jesper-friis:
> For me this PR looks reasonable. It is mostly renaming `prepare` to `as_jsonld`, and adding a possibility to add `other_entries` such as samples. I am still of the opinion that context should be put into the yaml file, and that we should use JSON-LD or [YAML-LD](https://json-ld.github.io/yaml-ld/spec/) directly, which would immeadiately add the `other_entries` functionality in a more "intuitive" way. That discussion is better in a different PR I think. If you @francescalb dont have any more comments, feel free to merge

I think that it is a good idea to be able to provide user-defined JSON-LD context. But it is for the advanced users. For normal users, we should make it as simple as possible. I think we should stick with JSON-LD (semantically), but represent it in YAML. YAML-LD is an extension of JSON-LD that I don't think we urgently need. It is still in a draft phase with no implementations that I am aware of.

PR #273, opened 2024-12-15
New TableDoc class providing a table interface for data documentation

jesper-friis:
# Description
Added the TableDoc class providing a table interface for data documentation. 

This PR builds on top of PR #272. Merge that one first.

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer


tripper/dataset/tabledoc.py
```diff
@@ -0,0 +1,68 @@
...
+    ):
+        self.header = header
+        self.data = data
+        self.type = type
+        self.prefixes = prefixes
+        self.context = context
+
+    def asdicts(self) -> "List[dict]":
+        """Return the table as a list of dicts."""
+        kw = {"_context": self.context} if self.context else {}
```
torhaugl, line 50: Should this be @context?
```suggestion
        kw = {"@context": self.context} if self.context else {}
```

torhaugl (changes requested):
Suggested some small documentation changes.

This PR adds a TableDoc class and addnested function with tests.
One issue is that we cannot add multiple of the same nested data (two distribution.downloadURL for instance). Can you add a warning for this in addnested?

Something like if d = {"a" : {"b" : 1}}, then a warning/error is issued
```python
d = {"a" : {"b" : 1}}
addnested(d, "a.b", 2)
### WARNING/ERROR, a.b already exists.
```

tripper/dataset/tabledoc.py
jesper-friis: > Should this be @context?

Either should work. Maybe "@context" is more readable. We support "_context" since "@context" cannot be provided as a keyword argument (unless using `**` unpacking, like we are doing here).

tests/dataset/test_tabledoc.py
```diff
@@ -0,0 +1,81 @@
...
+            ("ds:d1", "onto:T1", None, "ds:s1", "file:///data/d1.txt"),
+            ("ds:d2", "onto:T2", None, "ds:s1", "file:///data/d2.txt"),
+        ],
+        prefixes={
+            "onto": "http:/example.com/onto#",
+            "ds": "http:/example.com/datasets#",
+        },
+        # context={
+        #    "ds": "http:/example.com/datasets#",
+        # },
```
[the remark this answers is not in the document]
jesper-friis: I would like to keep this comment, since the intention is to uncomment it when we have a fully working implementation of the `context` keyword argument. Added a comment to clarify this

jesper-friis:
> ```python
> addnested(d, "a.b", 2)
> ```

The result of the above is

```python
{'a': {'b': [1, 2]}}
```

with the current implementation of `addnested()`. The purpose is to allow adding multiple of the same (nested) data.

PR #285, opened 2025-01-06
Re-applied fixes from testing datadoc on a use case for PINK.

jesper-friis:
# Description
Re-applied fixes from testing datadoc on a use case for PINK.

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/docs/tools/datadoc.md
+++ b/docs/tools/datadoc.md
@@ -282,10 +282,12 @@ options:
   -h, --help            show this help message and exit
   --type TYPE, -t TYPE  Either a resource type (ex: "dataset", "distribution",
                         ...) or the IRI of a class to limit the search to.
-  --criteria KEYWORD=VALUE [KEYWORD=VALUE ...], -c KEYWORD=VALUE [KEYWORD=VALUE ...]
-                        One of more additional matching criteria for resources
-                        to find. Only resources with the given KEYWORD and
-                        VALUE will be matched. The match is exact.
+  --criteria IRI=VALUE, -c IRI=VALUE
+                        Matching criteria for resources to find. The IRI may
+                        be written using a namespace prefix, like
+                        `tcterms:title="My title"`. Currently only exact
+                        matching is supported. This option can be given
+                        multiple times.
   --output FILENAME, -o FILENAME
                         Write matching output to the given file. The default
                         is to write to standard output.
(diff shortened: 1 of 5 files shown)


docs/tools/datadoc.md
francescalb, line 1: There is an extra ´with´ in line 51

francescalb, line 1: line 301: an --> can

francescalb, line 1: lien 305: options --> option

tripper/dataset/dataset.py
```diff
@@ -853,15 +853,19 @@ def get_partial_pipeline(
...
     Additional matching criterias can be specified by `kwargs`.
 
     Arguments:
         ts: Triplestore to search.
         type: Either a [resource type] (ex: "dataset", "distribution", ...)
             or the IRI of a class to limit the search to.
-        kwargs: Match criterias.
+        criterias: Match criterias. A dict of IRI, value pairs, where the
+            IRIs refer to data properties on the resource match. The IRIs
+            may use any prefix defined in `ts`, like `tcterms:title`.
```
francescalb, line 868: ``

francescalb, line 868:
```suggestion
            may use any prefix defined in `ts`. E.g. the prefix `dcterms` is understood, and can be used like `dcterms:title`.
```

francescalb:
Just a few typo-comments above.

However, I find the order og arguments to datadoc to be a bit unintuitive. Wouldn't it be better to have 
datadoc find/add/load and the non-positional arguments?

jesper-friis:
> However, I find the order og arguments to datadoc to be a bit unintuitive. Wouldn't it be better to have datadoc find/add/load and the non-positional arguments?

I do not disagree, but don't know how to improve. `datadoc` is using argparse sub-commands the standard way. The behaviour should be similar to git. Both for datadoc and git is the general syntax `git [GENERAL_OPTIONS...] subcommand [SUBCOMMAND_OPTIONS...]`. The only difference is that you seldom use the general options with git (since with git, all connection configurations are inferred from the working dir). But `git --git-dir=a.git --work-tree=b -C c status --short` is a possible and valid git command.

francescalb (approved):

PR #313, opened 2025-01-22
Add updateEndpoint to SPARQLwrapper

torhaugl:
# Description
Fixes the problem of having different URL for select and update queries, like in GraphDB.

```python
import tripper
from tripper import Literal

# CRASHES
#base_iri="http://GRAPHDB/repositories/TOPMAST"
#ts = tripper.Triplestore("sparqlwrapper", base_iri=base_iri)
#res = ts.add_triples([("<http://www.example.org/test1>", "<http://www.example.org/test3>", Literal("test3"))])
#res = ts.query("SELECT * WHERE { <http://www.example.org/test1> ?p ?o }")
#print(res)

#WORKS
select_iri="http://GRAPHDB/repositories/TOPMAST"
update_iri="http://GRAPHDB/repositories/TOPMAST/statements"

ts = tripper.Triplestore("sparqlwrapper", base_iri=update_iri)
res = ts.add_triples([("<http://www.example.org/test1>", "<http://www.example.org/test3>", Literal("test3"))])

ts = tripper.Triplestore("sparqlwrapper", base_iri=select_iri)
res = ts.query("SELECT * WHERE { <http://www.example.org/test1> ?p ?o }")
print(res)

#WITH THIS PR, update_iri keyword is added (optional)
ts = tripper.Triplestore("sparqlwrapper", base_iri=select_iri, update_iri=update_iri)
res = ts.add_triples([("<http://www.example.org/test1>", "<http://www.example.org/test3>", Literal("test3"))])
res = ts.query("SELECT * WHERE { <http://www.example.org/test1> ?p ?o }")
print(res)
```

## Type of change
- [X] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tripper/backends/sparqlwrapper.py
+++ b/tripper/backends/sparqlwrapper.py
@@ -1,9 +1,9 @@
 """Backend for SPARQLWrapper"""
 
+import warnings
 from typing import TYPE_CHECKING
 
 from tripper import Literal
-import warnings
 
 try:
     from SPARQLWrapper import GET, JSON, POST, RDFXML, SPARQLWrapper
--- a/tripper/backends/sparqlwrapper.py
+++ b/tripper/backends/sparqlwrapper.py
@@ -48,7 +48,9 @@ class SparqlwrapperStrategy:
         kwargs.pop(
             "database", None
         )  # database is not used in the SPARQLWrapper backend
-        self.sparql = SPARQLWrapper(endpoint=base_iri, updateEndpoint=update_iri, **kwargs)
+        self.sparql = SPARQLWrapper(
+            endpoint=base_iri, updateEndpoint=update_iri, **kwargs
+        )
         if username and password:
             self.sparql.setCredentials(username, password)
 
--- a/tripper/backends/sparqlwrapper.py
+++ b/tripper/backends/sparqlwrapper.py
@@ -50,9 +50,15 @@ class SparqlwrapperStrategy:
             endpoint=base_iri, updateEndpoint=update_iri, **kwargs
         )
         import warnings
-        if not self.sparql.isSparqlUpdateRequest() and self.sparql.updateEndpoint is None:
-            warnings.warn("The 'base_iri' is not a valid update endpoint. "
-                    "For updates it is necessary to give the 'update_iri'.")
+
+        if (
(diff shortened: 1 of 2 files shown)


torhaugl:
Attached is the error message.

My first idea to solve the error is to catch it, check if the response is "Missing parameter: query", then return an error message explaining that maybe the wrong endpoint is used. Perhaps, `Endpoint is missing parameter: query. This can happen when the endpoint does not support SPARQL updates.` However, I don't know if this would only work for GraphDB or also for other triplestores. They might return a different error message.

To automatically solve this with selecting the correct update endpoint, we need a specific implementation for GraphDB. The update endpoint is the select endpoint + '/statements' for GraphDB.

```
Traceback (most recent call last):
[...]
urllib.error.HTTPError: HTTP Error 400:

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]
SPARQLWrapper.SPARQLExceptions.QueryBadFormed: QueryBadFormed: A bad request has been sent to the endpoint: probably the SPARQL query is badly formed.

Response:
b'Missing parameter: query'
```

tripper/backends/sparqlwrapper.py
```diff
@@ -38,14 +40,17 @@ class SparqlwrapperStrategy:
...
         password: "Optional[str]" = None,
         **kwargs,
     ) -> None:
         kwargs.pop(
             "database", None
         )  # database is not used in the SPARQLWrapper backend
-        self.sparql = SPARQLWrapper(endpoint=base_iri, **kwargs)
+        self.sparql = SPARQLWrapper(
+            endpoint=base_iri, updateEndpoint=update_iri, **kwargs
+        )
```
francescalb, line 53:
```suggestion
        )
                import warnings
        if not self.sparql.isSparqlUpdateRequest() and self.sparql.updateEndpoint is None:
            warnings.warn("The 'base_iri' is not a valid update endpoint. "
                    "For updates it is necessary to give the 'update_iri'.")
```

How about something like this? Might be better to put it in the update methods, so that not send out warnings when we do not want to do updates.... And the import should be moved to the right place, of course.

francescalb (approved):

PR #340, opened 2025-02-19
Flb/expand iri in sparql query

francescalb:
# Description

## Type of change
- [ ] Bug fix and code cleanup
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer
This checklist should be used as a help for the reviewer.

- [ ] Is the change limited to one issue?
- [ ] Does this PR close the issue?
- [ ] Is the code easy to read and understand?
- [ ] Do all new feature have an accompanying new test?
- [ ] Has the documentation been updated as necessary?
- [ ] Is the code properly tested?

PR #342, opened 2025-02-19
Expand the IRI of @id in the dict returned by load_dict()

jesper-friis:
# Description
Expand the IRI of @id in the dict returned by load_dict(). 

This is a behaviour change. Is it really what we want? An alternative would be to add an option `expand_iri=False`. 

Applied on top of #340. Ensures that the sparql query works. 

## Type of change
- [x] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/backends/test_sparqlwrapper_graphdb.py
+++ b/tests/backends/test_sparqlwrapper_graphdb.py
@@ -133,12 +133,10 @@ ASK {
     # search for datasets in the graphDB
     datasets = search_iris(ts, type="dataset")
 
-    assert set(datasets) == set(
-        [
-            "https://onto-ns.com/datasets#our_nice_dataset",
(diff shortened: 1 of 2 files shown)


francescalb:
On further consideration I think we are tackling this from the wrong angle. We should not expand the iri that we fetch, as we do not know what the original prefix called for. However, we should expand the iri when we upload the documentation to the database.

jesper-friis:
> On further consideration I think we are tackling this from the wrong angle. We should not expand the iri that we fetch, as we do not know what the original prefix called for. However, we should expand the iri when we upload the documentation to the database.

I agree

jesper-friis:
I think that this PR is outdated. The dict returned by the `get_dict()` function now has all keys expanded to full IRIs. 

@francesca, ok for you to close this PR?

PR #347, opened 2025-02-27
Added test for Fuseki and fixed query to return TURTLE which works for both fuseki and graphdb

francescalb:
# Description
Added test for Fuseki and fixed query to return TURTLE which works for both fuseki and graphdb

## Type of change
- [ ] New feature
- [x] Documentation update
- [x] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/backends/test_sparqlwrapper_graphdb_fuseki.py
+++ b/tests/backends/test_sparqlwrapper_graphdb_fuseki.py
@@ -208,4 +208,10 @@ ASK {
     ts.bind("dataset", "https://onto-ns.com/datasets#")
     retreived_info_2 = load_dict(ts, f"dataset:{datasets[0].split('#')[-1]}")
     print(retreived_info_2)
-    assert retreived_info == retreived_info_2
+    assert retreived_info_2.creator == "Tripper-team"
+    assert (
+        retreived_info_2.title
+        == "This is a title of a completely invented dataset"
+    )
+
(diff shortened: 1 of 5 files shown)


tests/backends/test_sparqlwrapper_graphdb_fuseki.py
```diff
@@ -35,6 +36,26 @@ def graphdb_available():
         time.sleep(interval)
 
 
+def fuseki_available():
+    """
+    Help function that checks if the Fuseki instance is available
```
jesper-friis, line 41:
```suggestion
    Help function that returns True if the Fuseki instance is available
```

tests/backends/test_sparqlwrapper_graphdb_fuseki.py
```diff
@@ -35,6 +36,26 @@ def graphdb_available():
...
+    interval = 1  # seconds
+    start_time = time.time()
+    while True:
+        try:
+            response = requests.get(FUSEKI_CHECK_URL, timeout=timeout)
+            if response.status_code == 200:
+                return True
+        except requests.exceptions.RequestException:
+            pass
+
```
jesper-friis, line 53: Could maybe add a short sleep here to not consume so much bandwidth.

jesper-friis (approved):
Very, very nice to include run a Fuseki instance during CI testing.

Looks good. But I have not had time to test it...

PR #387, opened 2025-04-30
Improved control when storing to the triplestore

jesper-friis:
# Description
Added a `method` argument to `dataset.store()` that allows the user to control whether an existing IRI should be overwritten or merged. The default is to raise an exception.

Other changes:
* Renamed some functions with awkward names:
  - `save_dict() -> store()`
  - `load_dict() -> acquire()`
  - `search_iris() -> search()`
  
  Using old function names will work, but trigger a deprecation warning.
* Cleaned up documentation.

## Type of change
- [x] New feature
- [x] Documentation update
- [x] Testing

## Checklist for the reviewer

### DIFF
--- a/docs/datadoc/keywords.md
+++ b/docs/datadoc/keywords.md
@@ -14,13 +14,14 @@ The meaning of the columns are as follows:
 ## Special keywords (from JSON-LD)
 See the [JSON-LD specification] for more details.
 
-| Keyword    | Range         | Conformance | Definition                                                   | Usage note |
-|------------|---------------|-------------|--------------------------------------------------------------|------------|
-| [@id]      | IRI           | mandatory   | IRI identifying the resource to document.                    |            |
-| [@type]    | IRI           | recommended | Ontological class defining the class of a node.              |            |
-| [@context] | dict&#124list | optional    | Context defining namespace prefixes and additional keywords. |            |
-| [@base]    | list          | optional    | Base IRI against which relative IRIs are resolved.           |            |
-| [@graph]   | list          | optional    | Used for documenting multiple resources.                     |            |
+| Keyword    | Range         | Conformance | Definition                                                              | Usage note |
+|------------|---------------|-------------|-------------------------------------------------------------------------|------------|
+| [@id]      | IRI           | mandatory   | IRI identifying the resource to document.                               |            |
+| [@type]    | IRI           | recommended | Ontological class defining the class of a node.                         |            |
+| [@context] | dict&#124list | optional    | Context defining namespace prefixes and additional keywords.            |            |
+| [@base]    | namespace     | optional    | Base IRI against which relative IRIs are resolved.                      |            |
+| [@vocab]   | namespace     | optional    | Used to expand properties and values in @type with a common prefix IRI. |            |
+| [@graph]   | list          | optional    | Used for documenting multiple resources.                                |            |
 
 
 ## Properties on [Resource]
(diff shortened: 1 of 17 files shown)


docs/datadoc/keywords.md
```diff
@@ -11,6 +11,17 @@ The meaning of the columns are as follows:
...
 
+## Special keywords (from JSON-LD)
+See the [JSON-LD specification] for more details.
+
+| Keyword    | Range         | Conformance | Definition                                                   | Usage note |
+|------------|---------------|-------------|--------------------------------------------------------------|------------|
+| [@id]      | IRI           | mandatory   | IRI identifying the resource to document.                    |            |
+| [@type]    | IRI           | recommended | Ontological class defining the class of a node.              |            |
+| [@context] | dict&#124list | optional    | Context defining namespace prefixes and additional keywords. |            |
+| [@base]    | list          | optional    | Base IRI against which relative IRIs are resolved.           |            |
```
francescalb, line 22: is this namespace?

francescalb:
seems like load_dict is changed to aquire? there is also a get, what is the usage of that compared to acquire?

docs/datadoc/keywords.md
jesper-friis: Yes, it is a kind of default namespace, similar to `base` in turtle. It will be applied to keywords with no pre-defined mappings in the context.

jesper-friis: JSON-LD also has a `@vocab` keyword that may be even more useful.

tripper/datadoc/dataset.py
```diff
@@ -314,78 +313,128 @@ def addsuperclasses(d, cls):
...
+        keywords: Keywords object with additional keywords definitions.
+            If not provided, only default keywords are considered.
         prefixes: Dict with prefixes in addition to those included in the
             JSON-LD context.  Should map namespace prefixes to IRIs.
+        method: How to handle the case where `ts` already contains a document
+            with the same id as `source`. Possible values are:
+            - "overwrite": Remove existing documentation before storing.
+            - "retain": Raise an `IRIAlreadyExistsError` if the IRI of `source`
+              already exits in the triplestore.
+            - "merge": Merge `source` with existing documentation.
```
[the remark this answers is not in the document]
jesper-friis: "retain" - do you have a better name for this? Other possibilities could be "keep" or "raise", but I am not sure that they are better.

jesper-friis:
> seems like load_dict is changed to aquire? there is also a get, what is the usage of that compared to acquire?

Yes, `load_dict()` was changed to `acquire()` (and not `load()`, to avoid confusion with the `load()` function in the dataaccess module).

The `get()` function is more a helper function (like add()`) that is not made available in the `__init__.py` file. It is similar to `dict.get()` but returns always a list. Would you have preferred to rename `load_dict()` to `get()` instead of `acquire()`? If we want that, we should do it now. The existing `get()` can e.g. be changed to `_get()`.

tripper/datadoc/dataset.py
```diff
@@ -314,78 +314,135 @@ def addsuperclasses(d, cls):
...
+                raise IRIExistsError(f"Cannot overwrite existing IRI: {iri}")
+            elif method == "merge":
+                pass
+            elif method == "ignore":
+                warnings.warn(iri, category=IRIExistsWarning)
+                return doc
+            else:
+                raise ValueError(
+                    f"Invalid storage method: '{method}'. "
+                    "Should be one of: 'overwrite', 'retain' or 'merge'"
```
francescalb, line 391:
```suggestion
                    "Should be one of: 'overwrite', 'raise', 'ignore' or 'merge'"
```

francescalb (approved):
Looks OK. It seems that search and acquire and load are not presented in the the documenting-a-resource.md. I suggest this is added as an issue to complete the documentation of the documentation.

PR #429, opened 2025-09-16
Criteria in which either key or value are None

### LINKED ISSUE #425: Extended search functionality
https://github.com/EMMC-ASBL/tripper/blob/17c00937f6f1fa2e20755e87deb70a6f81ba906f/tripper/datadoc/dataset.py#L1399

The existing search function does not provide means to:

1. search for a value across all objects (case with missing predicate value)
2. search for existence of predicate (case with missing object value)

While 2. may not be needed (or could even be hard for users to understand), 1. is valuable for users since it facilitates sort of "full text search" where the user does not necessarily need to know the predicate value but knows the object value (e.g. title).

The API of the search call could be modified to support this in the following way: instead of a dictionary to pass in parameters, use a list of tuples `(predicate, object)`. Passing `None` for predicate value supports case 1 above. Passing `None` for object value supports case 2 above. Using this approach would also allow for cases where a user wants to search for multiple objects for single predicate (e.g. `dc:title=Foo | dc:title=Bar`).

Additionally, the existing `criterias` and `regex` parameters could be merged into one by using a list of tuples `(predicate, object, regex)` where `regex: bool = False` would signify whether to use regular expression search or not for a given criteria pair.

francescalb:
# Description
closes #425 

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/datadoc/test_dataset.py
+++ b/tests/datadoc/test_dataset.py
@@ -8,6 +8,9 @@ pytest.importorskip("yaml")
 pytest.importorskip("requests")
 pytest.importorskip("pyld")
 
+GRAPHDB_CHECK_URL = "http://localhost:7200/repositories"
+FUSEKI_CHECK_URL = "http://localhost:3030"
+
 
 def test__get_range():
     """Test _get_default_keywords()."""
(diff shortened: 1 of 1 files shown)


tests/datadoc/test_dataset.py
```diff
@@ -538,17 +538,78 @@ def test_datadoc():
         SEMDATA["SEM_cement_batch2/77600-23-001/77600-23-001_5kV_400x_m001"],
     }
 
+    # Filter on criterion, but without required value
+    assert set(
+        search(
+            ts,
+            criteria={"creator.name": ""},
```
jesper-friis, line 545: I think that the only None should be treated as a wildcard. The empty string should only match an empty string.

jesper-friis:
`criteria` and `regex` should be treated the same way

francescalb:
> Criteria with only the object specified returns no results. For example:
> 
> `[('http://pink-project.eu/carcinogenity#taid', 'TOX-252')]` returns one result (`http://pink-project.eu/carcinogenity#254`)
> 
> while
> 
> `[(None, 'TOX-252')]` returns an empty list.

Strange, this seems to work in the tests.

PR #537, opened 2026-04-24
Add documentation on how to make sure that new classes exist in context

francescalb:
# Description

When parsing several tables one after the other and they refer to each other, they must be added in the right order, and context must be updated in between. Added some documentation on this. 

Also added a small test

## Type of change
- [ ] New feature
- [x] Documentation update
- [ ] Testing

## Checklist for the reviewer


docs/datadoc/customisation.md
```diff
@@ -271,6 +271,46 @@ The `__init__()` method of the [TableDoc] class takes a `context` argument with
...
+prefixes = {"myns": "https://example.org/myns/"}
+
+# 1. Parse the table that defines classes
+classes_doc = TableDoc.parse_csv("classes.csv", context=context, prefixes=prefixes)
+
+# 2. Register the newly-parsed classes in the shared context
+update_context(classes_doc.asdicts(), context)
+
+# 3. Now parse tables that reference those classes — restrictions will be inferred correctly
+resources_doc = TableDoc.parse_csv("resources.csv", context=context, prefixes=prefixes)
```
jesper-friis, line 308: I think we should propose an easier and more user-friendly solution here. See suggestion in PR https://github.com/PINK-project/PINK-annotation-schema/pull/68.

In short, we can add the lists of dicts to a single variable

```python
>>> dicts = tabel1.asdicts() + tabel2.asdicts() + ...
```

and then use dataset.store() to store them into the triplestore in a single call. This allows store() to resolve all dependencies between the various tables.
A user-defined context can easily be added in this stage. 

```python
>>> jsonld = store(ts, dicts, context=context)
```

tripper/datadoc/dataset.py
```diff
@@ -557,6 +557,25 @@ def update_context(
     Currently this only adds classes defined in `source` to `context`.
 
     Returns the updated context.
+
+    Note:
+        When parsing multiple sources that cross-reference each other's
```
jesper-friis: This has to be updated when a more user-friendly method exists.

What about just add a link thre to the documentation. Then we have only one place to update when a more user-friendly solution is implemented.

tripper/datadoc/dataset.py
```diff
@@ -649,6 +668,18 @@ def infer_restriction_types(
 
         where `<N>` is a positive integer.
 
+    Note:
+        Class recognition is based solely on what is registered in the
```
jesper-friis: Same as above

tripper/datadoc/tabledoc.py
```diff
@@ -275,7 +275,17 @@ def parse_csv(
                 or an IRI to a class in an ontology.
             keywords: Keywords object with additional keywords definitions.
                 If not provided, only default keywords are considered.
-            context: Dict with user-defined JSON-LD context.
+            context: Context object or dict with user-defined JSON-LD
```
jesper-friis: and here

docs/datadoc/customisation.md
francescalb: This is easier and more userfriendly once you know what happens begind the scenes. Your suggestion can be added as an easier alternative, but I disagree with just writing that out.

docs/datadoc/customisation.md
```diff
@@ -271,6 +271,78 @@ The `__init__()` method of the [TableDoc] class takes a `context` argument with
...
+Technically, it is often easier to create one big dict with all the resources from all tables, and then call [store()] once on that dict. This way, all classes are visible to `infer_restriction_types()` when it processes the whole dict.
+
+```python
+from tripper.datadoc import get_context, TableDoc
+from tripper.datadoc.dataset import store
+
+context = get_context("https://example.org/context/", theme=None)
+prefixes = {"myns": "https://example.org/myns/"}
+
+# 1. Parse the table that defines classes
```
jesper-friis, line 332: This will be a main header in markdown. Numbered items are written without the initial hash

```suggestion
1. Parse the table that defines classes
```

docs/datadoc/customisation.md
```diff
@@ -271,6 +271,78 @@ The `__init__()` method of the [TableDoc] class takes a `context` argument with
...
+classes_doc = TableDoc.parse_csv("classes.csv", context=context, prefixes=prefixes)
+
+# 2. Parse the table that defines resources that reference those classes
+resources_doc = TableDoc.parse_csv("resources.csv", context=context, prefixes=prefixes)
+
+# 3. Create one big dict with all resources from all tables
+bigdict = classes_doc.asdicts() + resources_doc.asdicts()
```
jesper-friis: `bigdict` sounds like it is a dict. But it is a list of dicts. What about calling it `dicts`?

jesper-friis (commented):
Just two suggestions

docs/datadoc/customisation.md
jesper-friis: ok

docs/datadoc/customisation.md
francescalb: This is inside a code snippet

jesper-friis (approved):

PR #538, opened 2026-04-25
Avoid injecting default context

jesper-friis:
# Description
Ensure that default context doesn't leak into data documentation when it is not intended.
- `context.py`: We want to get away with keywords. For now, do not update context from keywords.
- `dataset.py`: Added theme argument to `told()` to make it possible to propage user-defined content with `theme=None`.
- `tabledoc.py`: Change default theme to None instead of "ddoc:datadoc"

This has the consequence that a few tests that were developed towards the old behaviour had to be updated:
- `test_datadoc_cli.py` now uses a copy of semdata-context.json (called semdata-context-ext.json) that includes the default context
- `test_tabledoc.py‎` commented out inferences that are no more included

Other changes:
- added convinient Context.type() method (used for testing)
- in keywords.yaml, added rdfs:Literal as range to dcterms:format -> this also updates the generated files
- fixed documentation bug in datadoc.clitool
- included virtualenvwrapper to pyproject.toml, since (for some strange reasons) pre-commit complains about missing virtualenvwrapper

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer


tests/datadoc/test_tabledoc.py
```diff
@@ -285,8 +285,8 @@ def test_csv_keywords():
     batch1 = td.asdicts()[0]
     assert batch1["@type"] == [
         "dcat:Dataset",
-        "dcat:Resource",
-        "emmo:EMMO_194e367c_9783_4bf5_96d0_9ad597d48d9a",
+        # "dcat:Resource",
+        # "emmo:EMMO_194e367c_9783_4bf5_96d0_9ad597d48d9a",
```
francescalb, line 289: Only comment out if there is a reason to keep the comment. Otherwise delete.

tripper/datadoc/context.py
```diff
@@ -57,15 +57,15 @@ def get_context(
...
             context = context.copy()
-        if keywords or theme:
-            kw = get_keywords(keywords=keywords, theme=theme)
-            context.add_context(kw.get_context())
+        # if keywords or theme:
+        #    # pylint: disable=import-outside-toplevel
+        #    from tripper.datadoc.keywords import get_keywords
+        #
+        #    kw = get_keywords(keywords=keywords, theme=theme)
+        #    context.add_context(kw.get_context())
```
francescalb, line 68: same a s above

tripper/datadoc/dataset.py
```diff
@@ -408,9 +410,9 @@ def store(
         source: Dict or list with the resource documentation to store.
         type: Type of documented resource.  Should be one of the resource types
             defined in `keywords`.
+        theme: IRI of one of more themes to load keywords for.
```
francescalb:
```suggestion
        theme: IRI of one or more themes to load keywords for.
```

francescalb (commented):
I find it difficult to identify the tests that cover the two different cases, i.e. when tripper context is included and when it is not. Making a test that shows the two different situations would make later debugging and understanding of the code easier. In particular the semdata-context-ext.json and the original are used in two completely different places. They should be tested side by side to show the difference.

tripper/datadoc/context.py
jesper-friis: This one I think we could keep. Here we are removing a potential usesul feature (that context is initialised from keywords). It is completely fine if we remove keywords altogether. But if we for some reasons choose to keep the keywords, then we probably want to reintroduce this feature. But it must be done in a way such that we avoid unintentional injection of the default context.

jesper-friis:
> I find it difficult to identify the tests that cover the two different cases, i.e. when tripper context is included and when it is not. Making a test that shows the two different situations would make later debugging and understanding of the code easier. In particular the semdata-context-ext.json and the original are used in two completely different places. They should be tested side by side to show the difference.

I agree. But I think it should be another PR. The intention with this PR was to fix such that the PINK documentation works. It fits well with the discussion about wheter we should get rid of the keywords module.

tripper/datadoc/context.py
francescalb: OK. I would prefer a comment on thing like that though.

francescalb:
> > I find it difficult to identify the tests that cover the two different cases, i.e. when tripper context is included and when it is not. Making a test that shows the two different situations would make later debugging and understanding of the code easier. In particular the semdata-context-ext.json and the original are used in two completely different places. They should be tested side by side to show the difference.
> 
> I agree. But I think it should be another PR. The intention with this PR was to fix such that the PINK documentation works. It fits well with the discussion about wheter we should get rid of the keywords module.

Again, a new PR "fixing" something but leaving the tests or documentation to a later time. It is really a waste of time to work like this.

francescalb (approved):

PR #539, opened 2026-04-26
Added parsing of excel and semantics to header braces in tabledoc

### LINKED ISSUE #529: TableDoc - add semantics to content of braces prepended to the header
With TableDoc, column headers can be followed by an optional brace `[...]`. Currently, Tripper does not define any semantics for the string within the brace.

The original motivation for introdusing the brace, was to have a way to allow duplicating columns, while still making the column headers unique, which is required by `pandas`. 
However, already at the initial design, it was foreseen that the brace could be useful for adding additional semantics to the column. For example, specifying a unit for the column.

The `dlite.table` module now uses the brace to identify the property that is documented in a given column.

It is time to define a general semantics for the brace, before we get to many diverging ways of using it.

Here we suggest the following general representation of of a column header:

    name[label?key1=value1&key2=value2]

where:
- `name`: is the column name that is mapped to a property via the JSON-LD context. The only formal requirement is that is cannot contain a begin brace ([), but it might be wise to be more strict.
- `label`: is a label for the column. It is used to make the column unique or to group related columns. The only formal requirement is that is cannot contain a question mark (?), but it probably wise to be more strict. A digit should be allowed.
- `key`: a key identifying an option for the column. Should be a valid C or Python identifier (regex: `[_a-zA-Z][_a-zA-Z0-9]*`).
- `value`: the value of a key. Should not contain (unescaped?)  ampersand (&) or end braces (]).

Some potential keys:
- **`unit`**: A unit symbol. All numbers in this column has this unit.
- **`sep`**: Separator character. A common user request when you have multiple values for a column, is to be able to provide multiple values in a single cell, instead of duplicating the column. This option makes it possible to specify a separator character that can be used in this column.

## Examples of use cases

Ensure that columns are unique (e.g. for use with pandas):

| @id | @type[1] | @type[2] |
| --- | --- | --- |

Grouping of columns (e.g. for DLite datamodels):

| @id | @type | datumName[1] | datumType[1] | datumName[2] | datumType[2] | 
| --- | --- | --- | --- | --- | --- |

Specifying unit:

| @id | @type | length[?unit=m] |
| --- | --- | --- |
| ex:my_length | emmo:Length | 3.2 |

Specifying a separator:

| @id | @type | keyword[?sep=;] |
| --- | --- | --- |
| ex:mydata | emmo:Dataset | geology;stone;cave |

jesper-friis:
# Description
Implements issue #529 

Also implemented parse_excel() method.

Closes #529 

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer


docs/datadoc/documenting-a-resource.md
```diff
@@ -174,7 +174,61 @@ Documenting as table
 The [TableDoc] class can be used to document multiple resources as rows in a table.
 
 The table must have a header row with defined keywords (either [predefined][predefined keywords] or provided with a custom context).
-Nested fields may be specified as dot-separated keywords. For example, the table
+Tripper allow to have multiple columns with the same keyword in the header.
```
francescalb:
```suggestion
Tripper allows multiple columns with the same keyword in the header.
```

docs/datadoc/documenting-a-resource.md
```diff
@@ -174,7 +174,61 @@ Documenting as table
...
+Nested fields may be specified as dot-separated keywords.
+Ex: `distribution.downloadURL`.
+
+The header keyword(s) may be followed by an optional square bracket with the following syntax (inspired by URL options):
+
+    name[label?var1=value1&var2=value2]
+
+where
+- `name`: is the column name that is mapped to a keyword (or set of dot-separated keywords) defined in the JSON-LD context.
+  The only formal requirement is that is cannot contain a begin brace ([), but it might be wise to be more strict.
```
francescalb:
```suggestion
  The only formal requirement is that it cannot contain a begin brace ([), but it might be wise to be more strict.
```

docs/datadoc/documenting-a-resource.md
```diff
@@ -174,7 +174,61 @@ Documenting as table
...
+- `label`: is a label for the column.
+  It is used to make the column unique or to group related columns.
+  The only formal requirement is that is cannot contain a question mark (?), but it probably wise to be more strict.
```
francescalb:
```suggestion
  The only formal requirement is that is cannot contain a question mark (?), but it is probably wise to be more strict.
```

docs/datadoc/documenting-a-resource.md
```diff
@@ -174,7 +174,61 @@ Documenting as table
...
+  A digit should be allowed.
+- `key`: a key identifying an option for the column.
+  Should be a valid C or Python identifier (regex: `[_a-zA-Z][_a-zA-Z0-9]*`).
+- `value`: the value of a key.
+  Should not contain (unescaped) ampersand (&) or end braces (]).
+
+Currently recognised keys:
+- **`unit`**: A unit symbol. All numbers in this column has this unit.
```
francescalb:
```suggestion
- **`unit`**: A unit symbol. All numbers in this column have this unit.
```

docs/datadoc/documenting-a-resource.md
francescalb:
```suggestion
corresponds to the following turtle representation:
```

tripper/datadoc/tabledoc.py
```diff
@@ -483,3 +566,70 @@ class dialect(csv.Dialect):
...
+    """Help class representing a column."""
+
+    # pylint: disable=too-few-public-methods
+
+    def __init__(self, head, context=None, strip=True):
+        # pylint: disable=line-too-long
+        """Initialise a column opject.
+
+        Arguments:
+            head: Column header. For the allowed syntax of the header label
```
francescalb:
```suggestion
            header: Column header. For the allowed syntax of the header label
```

jesper-friis: Hmm. Good naming is difficult. In `TableDoc.__init__()`, the "header" argument refer to the header of all columns in the whole reader row. Here we are only referring to the header of a single column, so I tried to find another name. I am open for suggestions, but I think that using "header" migth be confusing.

francescalb: I still think it makes sense to use header here, and it can be used in both places. The header for one column is one header.  Most correct would be to change header in __init__ to headers.

jesper-friis: Good suggestion. Renamed header->headers in `__init__()` and head->header here.

francescalb (approved):

PR #553, opened 2026-06-16
table_parse add header with prefix and as iris

francescalb:
# Description

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer


tests/datadoc/test_tabledoc.py
```diff
@@ -177,6 +204,9 @@ def test_csv():
 
     # Print serialised KB
     ts = Triplestore(backend="rdflib")
+    # from tripper.datadoc import store
+    # jsonld = store(ts, td.asdicts())
+    # print(jsonld)
```
francescalb:
```suggestion
```

.github/workflows/ci_check_dependencies.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       git_username: "TEAM 4.0[bot]"
       git_email: "Team4.0@SINTEF.no"
       target_branch: "ci/dependency-updates"
-      python_version: "3.9"
+      python_version: "3.13"
```
jesper-friis: Any reason for not going to v 3.14?

tripper/datadoc/tabledoc.py
```diff
@@ -614,11 +622,44 @@ def __init__(self, header, context=None, strip=True):
...
+
+        add_key = header
         if context and not leafname.startswith("@"):
-            df = context.getdef(leafname)
-            if "@type" in df and df["@type"] != "@id":
+            df = context.getdef(leafname, strict=False)
+
+            # If a full IRI or a prefixed name is used as header but the
+            # context does not define it, treat this as an error instead
+            # of silently accepting it.
```
jesper-friis, line 638: Being strict is probably a good idea. An alternative could be to treat it as an annotation property. 

Either way, we should treat prefixed and full IRI headers equally, as implemented here 👍

tripper/datadoc/tabledoc.py
```diff
@@ -614,11 +622,44 @@ def __init__(self, header, context=None, strip=True):
...
+                if leafname not in context.get_context_dict():
+                    entry = {"@id": df["@id"]}
+                    if "@type" in df:
+                        entry["@type"] = df["@type"]
+                    elif "@language" in df:
+                        entry["@language"] = df["@language"]
+                    context.add_context({leafname: entry})
+
+        # store key to use when adding nested values
+        self._add_key = add_key
```
jesper-friis: Is there any difference between `add_key` and `header`? 

If not, I think it is better to not introduce a new class attribute - especially since the name is not very intuitive.

tripper/datadoc/tabledoc.py
francescalb: In this way we keep the old functionality. Isn't that best?

tripper/datadoc/tabledoc.py
francescalb: I agree, I think it is a remnant of a lot of testing.

francescalb: removed

.github/workflows/ci_check_dependencies.yml
francescalb: Updated to 3.14

jesper-friis (approved):

PR #562, opened 2026-08-16
Simplify header syntax

jesper-friis:
# Description
Allowed to omit initial ?-sign in table header with no label. For example, allow the user to write "length[unit=m]" instead of "length[?unit=m]" in a table header.

Also updated the documentation accordingly.

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tripper/datadoc/tabledoc.py
+++ b/tripper/datadoc/tabledoc.py
@@ -601,7 +601,7 @@ class Column:
     # pylint: disable=too-few-public-methods
 
     def __init__(self, header, context=None, strip=True):
-        # pylint: disable=line-too-long
+        # pylint: disable=line-too-long,too-many-branches
         """Initialise a column opject.
 
         Arguments:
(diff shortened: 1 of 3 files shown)


tripper/datadoc/tabledoc.py
```diff
@@ -623,10 +623,18 @@ def __init__(self, header, context=None, strip=True):
 
         label = fields[0][2].split("?", 1)[0]
         spec = fields[-1][2].split("?", 1)
+        opts = spec[1] if len(spec) == 2 else ""
+
+        # Special rule. If `label` contain an equal sign (=), it is treated
+        # to be the first part of `opts`.
```
francescalb:
```suggestion
        # Special rule. If `label` contains an equal sign (=), it is treated
        # to be the first part of `opts`.
```

francescalb:
Looks OK, but please add an example with using both label and key, so that it becomes clearer that this is also allowed.

francescalb (approved):

PR #566, opened 2026-08-30
Corrected Namespace.__contains__() method

jesper-friis:
# Description
Corrected Namespace.__contains__() and added test.

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/test_namespace.py
+++ b/tests/test_namespace.py
@@ -73,7 +73,9 @@ def test_namespaces() -> None:
     assert "Child" in dir(FAM)
 
     # Test __contains__() method
-    assert "family#" in FAM
+    assert "type" in RDF
+    assert "nonexisting" in RDF  # This is the best we can do...
+    assert "Son" in FAM
     assert "foot" not in FAM
 
 
--- a/tripper/namespace.py
+++ b/tripper/namespace.py
@@ -274,8 +274,8 @@ class Namespace:
 
     def __contains__(self, name):
         if self._iris is None:
-            return True
-        if self._iris == {}:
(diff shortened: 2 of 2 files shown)


tests/test_namespace.py
```diff
@@ -73,7 +73,9 @@ def test_namespaces() -> None:
     assert "Child" in dir(FAM)
 
     # Test __contains__() method
-    assert "family#" in FAM
+    assert "type" in RDF
+    assert "nonexisting" in RDF  # This is the best we can do...
```
francescalb, line 77: I do not understand this. "nonexisting" is not present in https://www.w3.org/1999/02/22-rdf-syntax-ns, and should not be in the Namespace, I think.

jesper-friis: Since the RDF namespace is not created with `check=False`, the namespace object is not doing any checking on the validity of the IRI. That means that if you write `RDF.nonexisting` it will return `https://www.w3.org/1999/02/22-rdf-syntax-ns#nonexisting`.

Hence, it behaves like "nonexisting" exists in the `RDF` namespace. Therefore, we consider `"nonexisting" in RDF` to be true. 

This is not the case for `FAM`, which has been created with `check=True`.

francescalb: OK, this needs to be commented in the test, otherwise it is very easily misunderstood later. Otherwise fine

francescalb, line 77:
```suggestion
    # This will return True since the RDF Namespace is created with check=False
    assert "nonexisting" in RDF
```

PR #567, opened 2026-08-30
Don't convert None returned by ts.query to string

jesper-friis:
# Description
Don't convert None returned by ts.query to string

## Type of change
- [ ] New feature
- [ ] Documentation update
- [ ] Testing

## Checklist for the reviewer

### DIFF
--- a/tests/test_sparql.py
+++ b/tests/test_sparql.py
@@ -5,7 +5,6 @@
 import pytest
 
 
-# if True:
 def test_sparql_select():
     """Test SPARQL SELECT query."""
     pytest.importorskip("rdflib")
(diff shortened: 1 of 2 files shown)


francescalb:
seems to be implemented for the rdflib backend only. How about the other bakends? (rdflib is just intended for local testing, so proper graph backends are much more important.)

jesper-friis:
> seems to be implemented for the rdflib backend only. How about the other bakends? (rdflib is just intended for local testing, so proper graph backends are much more important.)

True. It is a bug in the rdflib backend. I have not checked other backends...

I will suggest that we fix eventual similar bugs in other backends in separate PRs...

francescalb (approved):

PR #570, opened 2026-09-01
Allow tabledoc to separate columns on multiple characters

jesper-friis:
Small PR that allows Tabledoc to separate columns on multiple characters.

The PR allows different characters to be used for splitting a column. 

### Potential surprising side effect
A column with header "keyword[?sep=, ]" would split the cells "kw1, kw2" and "kw1, , kw2" the same way. 

Hence, the current implementation does not allow to include empty intermediate fields in the list (although empty initial or final field is possible, e.g. with ", kw1, kw2").

Is this behaviour so unwanted that we should do something about it? Intuitively, you would expect that consecutive spaces are threated as one, while all other characters should be treated as they are given. That would lead to quite complex rules for creating the correct regexp. Examples:

* `[sep= ]` --> regexp: `" +"`
* `[sep=,]` --> regexp: `"[,]"` (bracket is not needed in this case, but added to protect against regexp injection)
* `[sep=, ]` --> regexp: `"[,] *"`
* `[sep=,; ]` --> regexp: `"[,;] *"`
* `[sep=, ;]` --> ??? Should this be translated to `"[,;] *"` or something different?

One possible rule could be that all consecutive non-space characters in `sep` that comes before or after the space are placed in a shared bracket. Hence `[sep=, ;]` --> regexp: `"[,] *[;]"`


francescalb:
We have discussed this, and decided that we allow only one delimiter per column. Note that the delimiter can be specified for each column. It is important to not make it so flexible that it becomes difficult to handle. I think it is fair to expect the producers of the tables to be able to use only one delimiter per column.

